PR #31, opened 2020-03-12
Add template module.

maltoe:
Screenshot 2020-03-13 at 00 12 53

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Added
 
 - HTML source and header and footer template accept iolists now
+- New `Template` module contains basic CSS skeleton to easily dimension & layout pages (header & footer margins)
 
 ## [0.2.0] - 2020-03-06
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ end
 
 ## Usage
 
-### Example
+### Main API
 
 ```elixir
 ChromicPDF.print_to_pdfa(
@@ -106,6 +106,34 @@ ChromicPDF.print_to_pdfa(
   # Output path.
   output: "test.pdf"
 )
+```
+
+### Sample boilerplate for a template
+
+This more sophisticated example shows how to use `Phoenix.View` to move HTML markup into template
+files.
+
+```elixir
+defmodule MyApp.InvoiceView do
+  use Phoenix.View,
+    root: "lib/my_app/path/to/pdf_templates",
+    namespace: MyApp
+
+  def print_to_pdf(assigns) do
+    [content: content(assigns)]
+    |> ChromicPDF.Template.source_and_options()
+    |> ChromicPDF.print_to_pdf()
+  end
+
+  @style File.read!("../path/to/assets/css/invoice.css")
+
+  defp content(assigns) do
+    [
+      @style,
+      Phoenix.HTML.safe_to_string(render("content.html", assigns))
+    ]
+  end
+end
 ```
 
 ## Development
--- a/lib/chromic_pdf/processor.ex
+++ b/lib/chromic_pdf/processor.ex
@@ -9,6 +9,7 @@ defmodule ChromicPDF.Processor do
   @type blob :: iodata()
 
   @type source :: {:url, url()} | {:html, blob()}
+  @type source_and_options :: %{source: source(), opts: [pdf_option()]}
   @type return :: :ok | {:ok, binary()}
 
   @type output_option :: {:output, binary()} | {:output, function()}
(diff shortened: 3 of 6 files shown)


lib/chromic_pdf/pdf/template.ex
```diff
@@ -0,0 +1,172 @@
...
+      header_height: Keyword.get(opts, :header_height, "0"),
+      header_font_size: Keyword.get(opts, :header_font_size, "10pt"),
+      footer_height: Keyword.get(opts, :footer_height, "0"),
+      footer_font_size: Keyword.get(opts, :footer_font_size, "10pt")
+    ]
+
+    do_render_styles(assigns)
+  end
+
+  EEx.function_from_string(:defp, :do_render_styles, @styles, [:assigns])
```
nickgnd: 💡

lib/chromic_pdf/pdf/template.ex
```diff
@@ -0,0 +1,172 @@
+defmodule ChromicPDF.Template do
+  @moduledoc """
+  This module contains basic styles and options that make it simple to make the HTML content fully
+  cover a given page.
+  """
```
nickgnd: This does not sound fully clear to me.
I would probably mention that this module allows you to create a reusable template with custom content such header, footer, style and even content.

Something like:

"Use this module to create your HTML template with a custom header, footer, content and style.

or

"This module exposes helper methods to create your own HTML page template, including header, footer, content and style."


Feel free to ignore my comment if it is incorrect 😉

lib/chromic_pdf/pdf/template.ex
```diff
@@ -0,0 +1,172 @@
...
+
+  require EEx
+
+  @type blob :: binary()
+
+  @type content_option ::
+          {:html, blob()}
```
nickgnd: Why did you choose `:html` for this variable name?
I fear that it can lead to confusion. `:header` and `:footer` are valid HTML too.
What about `:body`, `:page_content`, or simply `:content`?

lib/chromic_pdf/pdf/template.ex
```diff
@@ -0,0 +1,172 @@
...
+  <h2>User Agent</h2>
+  <pre id="user-agent"></pre>
+
+  <script type="text/javascript">
+  window.onload = function() {
+    var browser, userAgent = navigator.userAgent;
+    document.getElementById('user-agent').innerHTML = userAgent;
+  };
+  </script>
+  """
```
nickgnd: Cool!

lib/chromic_pdf/pdf/template.ex
```diff
@@ -0,0 +1,172 @@
...
+      height: Keyword.get(opts, :height, "215.9mm"),
+      width: Keyword.get(opts, :width, "279.4mm"),
+      header_height: Keyword.get(opts, :header_height, "0"),
+      header_font_size: Keyword.get(opts, :header_font_size, "10pt"),
+      footer_height: Keyword.get(opts, :footer_height, "0"),
+      footer_font_size: Keyword.get(opts, :footer_font_size, "10pt")
+    ]
+
+    do_render_styles(assigns)
+  end
```
nickgnd: Mmhh 🤔 
This seems pretty strict to me. Basically you can customize only height and font size of both header and footer, do I understand well? What if you have a more convoluted footer with different texts?
Do you need to style it inline?

lib/chromic_pdf/pdf/template.ex
maltoe: 🤔 I wanted to align it with [`ChromicPDF.print_to_pdf/2`](https://github.com/bitcrowd/chromic_pdf/blob/0f0064c5c84c3ff6b055c873ffd885f2a3ebfc5e/lib/chromic_pdf/supervisor.ex#L114) where I also called it `:html`.

But you're right, it is indeed confusing and might even be a bit misleading in this case, as the param should only be body content markup, and don't have the enclosing `<html></html>` tags... Chromium has this print template in which they call it `content`, so I might go with this (and change the `:html` in `print_to_pdf/2`, as well) https://cs.chromium.org/chromium/src/components/printing/resources/print_header_footer_template_page.html?g=0&l=125

lib/chromic_pdf/pdf/template.ex
maltoe: :) hmm, "create your own HTML page template" -> I would guess there is more to it than this file, so I'd like to keep it closer to "these are some helpers that help configure the page layout and margins" or so. Nevertheless, it's unclear, let me rephrase it 👍

lib/chromic_pdf/pdf/template.ex
maltoe: Yes, the idea here, or the current functionality of `Template`, is only to give you configuration + absolutely minimum styles to easily create a page that has visible & consistent header & footer, and no weird other paddings/margins anywhere. It's surprisingly difficult to do, at first :)

All client styles and more layout styles need to be passed to `:html`, `:header`, `:footer` as `<style>` tags. I'll work on the documentation a bit.

thanks for your review 💚

lib/chromic_pdf/pdf/template.ex
maltoe: hmm realized the param is actually called `html` in devtools, so going to keep it htere. But can still change it in the template, because why not 🤷‍♂️

https://chromedevtools.github.io/devtools-protocol/tot/Page#method-setDocumentContent

maltoe:
@nickgnd updated the docs a bit, better like this?

PR #38, opened 2020-04-09
Restart session after a number of prints.

maltoe:
* This PR adds logic to `Session` that makes it automatically close and re-create its Target
* This in turn causes Chrome to create a new OS process (and kill the old one), keeping its overall memory footprint in check.

### DIFF
--- a/lib/chromic_pdf/pdf/session.ex
+++ b/lib/chromic_pdf/pdf/session.ex
@@ -2,7 +2,9 @@ defmodule ChromicPDF.Session do
   @moduledoc false
 
   use GenServer
-  alias ChromicPDF.{Browser, SpawnSession}
+  alias ChromicPDF.{Browser, CloseTarget, CreateTarget, SpawnSession}
+
+  @default_max_session_uses 1000
 
   # ------------- API ----------------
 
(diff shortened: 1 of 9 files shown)


lib/chromic_pdf/pdf/session.ex
```diff
@@ -21,25 +23,67 @@ defmodule ChromicPDF.Session do
...
-    %{browser: browser, session_id: session_id} = state
+  def handle_continue(:restart_session, state) do
+    %{opts: opts, target_id: target_id} = state
+
+    {:ok, true} =
+      [targetId: target_id]
+      |> CloseTarget.new()
+      |> run_protocol_through_browser(opts)
+
+    {:noreply, start_session(opts)}
```
andreasknoepfle, line 56: Nice solution with `handle_continue` 👏

lib/chromic_pdf/pdf/session.ex
```diff
@@ -21,25 +23,67 @@ defmodule ChromicPDF.Session do
...
+  defp increase_session_age(state) do
+    %{opts: opts, session_age: old_session_age} = state
 
-    protocol = protocol_mod.new(session_id, params)
-    response = Browser.run_protocol(browser, protocol)
+    session_age = old_session_age + 1
+    max_session_age = Keyword.get(opts, :max_session_age, @default_max_session_age)
 
-    {:reply, response, state}
+    {session_age >= max_session_age, %{state | session_age: session_age}}
```
andreasknoepfle, line 87: age sounds like time to me. Maybe `max_session_uses`?

test/integration/session_restart_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  @ps_cmd :"ps ax | grep -v grep | grep -i Chrom"
+  @wait_ms 500
+
+  defp chrome_pids do
+    @ps_cmd
+    |> :os.cmd()
+    |> to_string()
+    |> String.trim()
+    |> String.split("\n")
+    |> Enum.map(fn x -> Regex.named_captures(~r/[^\d]*(?<pid>[\d]+)/, x)["pid"] end)
```
andreasknoepfle, line 13: maybe some comment on how you achieved os compatibility here could save later generations of devs :)

test/integration/session_restart_test.exs
```diff
@@ -0,0 +1,50 @@
...
+      assert pids0 == pids1
+
+      print_and_wait()
+      pids2 = chrome_pids()
+
+      # After the second print operation, we expect the Session to have restarted its target
+      # process, so exactly one pid should have changed.
+      assert length(pids1) == length(pids2)
+      assert assert(length(pids1 -- pids2)) == 1
+      assert assert(length(pids2 -- pids1)) == 1
```
andreasknoepfle, line 47: Nice test 💚

lib/chromic_pdf/pdf/session.ex
andreasknoepfle: or `max_session_commands`?

andreasknoepfle: `max_session_prints` :D ?

maltoe: I like `max_session_uses` 👍

PR #46, opened 2020-07-10
Change Template options around.

maltoe:
* Use `paperWidth` and `paperHeight` after all, as `@page` alone does
not seem to be effective.
* Add `zoom` rule to scale header & footer the same as the content.
* Print background colors by default.

|without zoom|with zoom|
|-|-|
|Screenshot 2020-07-10 at 09 40 07|Screenshot 2020-07-10 at 09 40 28|

<details>
  <summary>Test code for above</summary>

```
defmodule ChromicPDF.Test do
  def test do
    styles = """
    <style>
      .test {
        margin-top: 1mm;
        height: 18mm;
        width: 30mm;
        background-color: green;
      }
    </style>
    """

    html = [styles, ~s(<div class="test"></div>)]

    [
      content: html,
      header: html,
      footer: html,
      header_height: "20mm",
      footer_height: "20mm",
      size: :a4
    ]
    |> ChromicPDF.Template.source_and_options()
    |> ChromicPDF.print_to_pdf(output: "file.pdf")
  end
end
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 - Removed the `:width` and `:height` options from `Template.styles/1` and
   `Template.source_and_options/1` as it turns out that Chrome does not pay attention to `@page`
-  dimensions and instead still sets the size of the produced PDF to US letter. Since I could not
-  figure out a way to set the PDF size besides using the `paperWidth` and `paperHeight` options,
-  I resided to moving to a `:size` option instead that accepts names like `:a4`, `:letter`, and
-  tuples of `{<width>, <height>}` in inches. These are then passed to `paperWidth` and
-  `paperHeight`.
+  dimensions and instead still sets the size of the produced PDF to US letter. Since it does not
+  seem possible to the PDF size in Chrome headless besides using the `paperWidth` and
+  `paperHeight` options, modved to a `:size` option instead that accepts names like `:a4`,
+  `:us_letter`, and tuples of `{<width>, <height>}` in inches. These are then passed to
+  `paperWidth` and `paperHeight`.
 - Ditched the `preferCssPageSize` option from `Template.source_and_options/1` as it did not seem
   to have any effect. See above.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   `Template.source_and_options/1` as it turns out that Chrome does not pay attention to `@page`
   dimensions and instead still sets the size of the produced PDF to US letter. Since it does not
   seem possible to the PDF size in Chrome headless besides using the `paperWidth` and
-  `paperHeight` options, modved to a `:size` option instead that accepts names like `:a4`,
+  `paperHeight` options, moved to a `:size` option instead that accepts names like `:a4`,
   `:us_letter`, and tuples of `{<width>, <height>}` in inches. These are then passed to
   `paperWidth` and `paperHeight`.
 - Ditched the `preferCssPageSize` option from `Template.source_and_options/1` as it did not seem
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/chromic_pdf/template.ex [hunk 1]
```diff
@@ -6,6 +6,14 @@ defmodule ChromicPDF.Template do
   set of page sizing options. Using this module is entirely optional, but perhaps can help to
   avoid some common pitfalls arising from the slightly unintuitive and sometimes conflicting
   behaviour of `printToPDF` options and `@page` CSS styles in Chrome.
+
+  One particularly cumbersome detail is that Chrome in headless mode does not correctly interpret
+  the `@page` CSS rule to configure the page dimensions. Resulting PDF files will always be in
+  US-letter format unless configured differently with the `paperWidth` and `paperHeight` options.
+  In my experience, results will be best if the `@page` rule aligns with the values passed to
```
andreasknoepfle, line 13:
```suggestion
  Experience has shown, that results will be best if the `@page` rule aligns with the values passed to
```

lib/chromic_pdf/template.ex
```diff
@@ -17,14 +25,20 @@ defmodule ChromicPDF.Template do
...
-          | {:height, binary()}
+          {:size, paper_size()}
           | {:header_height, binary()}
           | {:header_font_size, binary()}
           | {:footer_height, binary()}
           | {:footer_font_size, binary()}
 
+  @paper_sizes_in_inch %{
+    a4: {8.3, 11.7},
+    letter: {8.5, 11.0}
```
andreasknoepfle, line 39: it letter clear enough? Should this maybe be `us_letter` ?

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
 
+### Changed
+
+- Removed the `:width` and `:height` options from `Template.styles/1` and
+  `Template.source_and_options/1` as it turns out that Chrome does not pay attention to `@page`
+  dimensions and instead still sets the size of the produced PDF to US letter. Since I could not
+  figure out a way to set the PDF size besides using the `paperWidth` and `paperHeight` options,
+  I resided to moving to a `:size` option instead that accepts names like `:a4`, `:letter`, and
+  tuples of `{<width>, <height>}` in inches. These are then passed to `paperWidth` and
+  `paperHeight`.
```
andreasknoepfle, line 11: Possibly we could remove the `I` here as well and us a more neutral language. But since it is the changelogs, 🤷

lib/chromic_pdf/template.ex
maltoe: I think from the context it should be clear, but I changed it nonetheless.

From [Wikipedia: en/Paper_size](https://en.wikipedia.org/wiki/Paper_size)

> By 1977, A4 was the standard letter format in 88 of 148 countries. Today the standard has been adopted by all countries in the world except the United States and Canada. In Mexico, Costa Rica, Colombia, Venezuela, Chile, and the Philippines, the US letter format is still in common use, despite their official adoption of the ISO standard.

CHANGELOG.md
```diff
@@ -4,11 +4,11 @@
...
 - Removed the `:width` and `:height` options from `Template.styles/1` and
   `Template.source_and_options/1` as it turns out that Chrome does not pay attention to `@page`
-  dimensions and instead still sets the size of the produced PDF to US letter. Since I could not
-  figure out a way to set the PDF size besides using the `paperWidth` and `paperHeight` options,
-  I resided to moving to a `:size` option instead that accepts names like `:a4`, `:letter`, and
-  tuples of `{<width>, <height>}` in inches. These are then passed to `paperWidth` and
-  `paperHeight`.
+  dimensions and instead still sets the size of the produced PDF to US letter. Since it does not
+  seem possible to the PDF size in Chrome headless besides using the `paperWidth` and
+  `paperHeight` options, modved to a `:size` option instead that accepts names like `:a4`,
```
andreasknoepfle, line 9:
```suggestion
  `paperHeight` options, moved to a `:size` option instead that accepts names like `:a4`,
```

Issue #54, opened 2020-10-05
When using chromic_pdf in production, chromium is often crashing


skeisman:
chromic_pdf is working in production but a few times a day we get errors like below. We are using chromium. We are not using sandbox mode and none of the page assets are loaded via network. They are all embeded via base64. If you have any insights or debugging tips, let me know.

GenServer ChromicPDF.Browser terminating
** (FunctionClauseError) no function clause matching in ChromicPDF.Browser.handle_info/2
    (chromic_pdf 0.5.2) lib/chromic_pdf/pdf/browser.ex:74: ChromicPDF.Browser.handle_info({:EXIT, #PID<0.7925.0>, :chrome_has_crashed}, %{dispatch: #Function<1.51776475/1 in ChromicPDF.Browser.init/1>, protocols: []})
    (stdlib 3.13) gen_server.erl:680: :gen_server.try_dispatch/4
    (stdlib 3.13) gen_server.erl:756: :gen_server.handle_msg/6
    (stdlib 3.13) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
Last message: {:EXIT, #PID<0.7925.0>, :chrome_has_crashed}

maltoe:
Hi @skeisman,

sorry, unfortunately we don't have much insight for you.

1) the error message indicates that we're missing a `handle_info` clause in the `Browser` module, which should probably just bubble up the error by returning `{:stop, ...}` so the supervisor can restart both. However, I suppose the supervision works well for you and Chrome is automatically restarted?

2) Chrome probably spits some error message to stderr, but we're discarding it (see [here](https://github.com/bitcrowd/chromic_pdf/blob/f709f55a43db9152d84ce015e9fda922e0ed6899/lib/chromic_pdf/pdf/chrome_impl.ex#L34), we need to do this as it otherwise messes up the devtools protocol messages) - you could try to instead redirect it to a logfile and see if that gives you some hints?

maltoe:
Also, monitoring the chromium process would be great (though a bit difficult to do), for instance it could be that chromium has some unhealthy memory usage pattern and is killed by the OOM killer or so. You could log the pid when Chrome is started (e.g. [here](https://github.com/bitcrowd/chromic_pdf/blob/f709f55a43db9152d84ce015e9fda922e0ed6899/lib/chromic_pdf/pdf/chrome_impl.ex#L9)) and somehow trace this process externally... just an idea though, I know this is a lot to ask from you :)

[...]

PR #60, opened 2020-11-06
Don't trap exit in Browser.

maltoe:
We either need to handle the `{:EXIT, _pid, :chrome_has_crashed}`
message or alternatively just not trap the exit, which will make the
`Browser` process die alongside its linked `Connection` process.

Addresses #54.

See http://erlang.org/doc/man/erlang.html#process_flag-2


maltoe:
nevermind this...

i'm finally beginning to see

PR #81, opened 2020-12-26
[#80] Wait for Page.loadEventFired after setDocumentContent.

### LINKED ISSUE #80: Images using an absolute path (both file or remote URL) don't work with HTML source
When specifying the HTML content directly to the `print_to_pdf`, the rendered PDF doesn't contain the image. It works ok when specifying the image through base64. Going through the code, it seems that when in `:html` mode, there is no wait for the `Page.frameStoppedLoading` notification like it is for the `:url` mode.  https://github.com/bitcrowd/chromic_pdf/blob/29a23ac70ba5b4fb659bd28d4396e69b90ab949b/lib%2Fchromic_pdf%2Fpdf%2Fprotocols%2Fprint_to_pdf.ex#L13-L34

Reproduction: 

```ex
path = System.tmp_dir() |> Path.join("output.pdf")

result = ChromicPDF.Template.source_and_options(content: content()) |> ChromicPDF.print_to_pdf(output: path)
IO.inspect(result)

defp content() do
  :erlang.iolist_to_binary(["<img src=\"https://homepages.cae.wisc.edu/~ece533/images/peppers.png\">"])
end
```

maltoe:
This patch changes the `PrintToPDF` protocol to wait for Chrome's
`Page.loadEventFired` notification after `setDocumentContent` instead of
only for the response of the call. This gives Chrome time to fetch external
resources (``, `<script>`, etc.) from URLs before printing the page.

* Change the handling of `:await`-type protocol steps to allow multiple
  successive responses/notifications to arrive out-of-order and wait for
  all of them before continuing with the next `:call` step. This makes
  it possible to wait for both the response of the `setDocumentContent`
  call and the `loadEventFired` notification.
* Change the "blanking" of the page after printing to navigate to a local
  file (`priv/blank.html`) instead of `about:blank`, as otherwise Chrome
  refuses to fetch any `file:///` URLs contained within the HTML content.
* Navigate to the blank file before first print job in `SpawnSession`.

Fixes #80.

### DIFF
--- a/lib/chromic_pdf/pdf/protocol_macros.ex
+++ b/lib/chromic_pdf/pdf/protocol_macros.ex
@@ -64,7 +64,6 @@ defmodule ChromicPDF.ProtocolMacros do
         )
       end
 
-      defp skip_branch([], acc, _opts), do: acc
       defp skip_branch([:end | rest], acc, opts), do: do_build_steps(rest, acc, opts)
       defp skip_branch([_skipped | rest], acc, opts), do: skip_branch(rest, acc, opts)
     end
--- a/lib/chromic_pdf/pdf/protocols/spawn_session.ex
+++ b/lib/chromic_pdf/pdf/protocols/spawn_session.ex
@@ -1,9 +1,14 @@
 defmodule ChromicPDF.SpawnSession do
   @moduledoc false
 
+  import ChromicPDF.Utils, only: [priv_asset: 1]
   import ChromicPDF.ProtocolMacros
 
   @version Mix.Project.config()[:version]
+
+  def blank_url do
+    "file://#{priv_asset("blank.html")}"
+  end
 
   steps do
     call(:create_browser_context, "Target.createBrowserContext", [], %{"disposeOnDetach" => true})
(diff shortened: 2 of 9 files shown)


lib/chromic_pdf/pdf/protocol_macros.ex
```diff
@@ -64,7 +64,6 @@ defmodule ChromicPDF.ProtocolMacros do
         )
       end
 
-      defp skip_branch([], acc, _opts), do: acc
```
maltoe, line 67: drive-by, this clause should never happen

lib/chromic_pdf/pdf/protocols/spawn_session.ex
```diff
@@ -1,10 +1,15 @@
 defmodule ChromicPDF.SpawnSession do
   @moduledoc false
 
+  import ChromicPDF.Utils, only: [priv_asset: 1]
   import ChromicPDF.ProtocolMacros
 
   @version Mix.Project.config()[:version]
 
+  def blank_url do
+    "file://#{priv_asset("blank.html")}"
```
maltoe, line 10: I guess this will break on Windows 🤷‍♂️

PR #84, opened 2021-01-06
[83] Add global timeout option for browser interaction.

### LINKED ISSUE #83: Generating Large PDFs causes Timeout
Hello,

I'm attempting to generate a large pdf (100 pages) however the GenServer times out after only 5 (!) seconds. Is there a way to override this? I don't see a supervisor option to do so, and I can't seem to track down the timeout in the source.

Regards,
Dakora

maltoe:
This patch adds a new global `timeout` option that can be passed via the
session pool configuration and which allows to configure the GenServer
timeout (essentially how long we wait for "protocols" to complete).

Fixes #83.

### DIFF
--- a/lib/chromic_pdf/pdf/browser/channel.ex
+++ b/lib/chromic_pdf/pdf/browser/channel.ex
@@ -16,9 +16,24 @@ defmodule ChromicPDF.Browser.Channel do
     GenServer.start_link(__MODULE__, args)
   end
 
-  @spec run_protocol(pid(), Protocol.t()) :: {:ok, any()} | {:error, term()}
-  def run_protocol(pid, %Protocol{} = protocol) do
-    GenServer.call(pid, {:run_protocol, protocol})
+  @spec run_protocol(pid(), Protocol.t(), timeout()) :: {:ok, any()} | {:error, term()}
+  def run_protocol(pid, %Protocol{} = protocol, timeout) do
+    GenServer.call(pid, {:run_protocol, protocol}, timeout)
+  catch
+    :exit, {:timeout, {GenServer, :call, [_pid, {:run_protocol, _protocol}, _timeout]}} ->
+      raise("""
+      Timeout in Channel.run_protocol/3!
+
+      The underlying GenServer.call/3 exited with a timeout. This happens when the browser was
+      not able to complete the current operation (= PDF print job) within the configured
+      #{timeout} milliseconds.
+
+      If you are printing large PDFs and expect long processing times, please consult the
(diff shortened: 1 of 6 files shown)


lib/chromic_pdf/pdf/browser/session_pool.ex
```diff
@@ -64,10 +70,15 @@ defmodule ChromicPDF.Browser.SessionPool do
...
        browser: browser,
        spawn_protocol: spawn_protocol(args),
-       max_session_uses: max_session_uses(args)
+       max_session_uses: max_session_uses(args),
+       timeout: timeout(args)
      }}
   end
 
+  defp timeout(args) do
+    get_in(args, [:session_pool, :timeout]) || @default_timeout
```
lwassermann, line 79: Where is this imported? I wanted to make sure this is the Keyword.get_in not the Dict.get_in.

lib/chromic_pdf/pdf/browser/channel.ex
```diff
@@ -16,9 +16,24 @@ defmodule ChromicPDF.Browser.Channel do
...
+    :exit, {:timeout, {GenServer, :call, [_pid, {:run_protocol, _protocol}, _timeout]}} ->
+      raise("""
+      Timeout in Channel.run_protocol/3!
+
+      The underlying GenServer.call/3 exited with a timeout. This happens when the browser was
+      not able to complete the current operation (= PDF print job) within the configured
+      #{timeout} milliseconds.
+
+      If you are printing large PDFs and expect long processing times, please consult the
+      documentation for the `timeout` option of the session pool.
```
lwassermann, line 32: Is the documentation generated from the types? Otherwise, where is the documentation for the timeout option?

lwassermann (approved):
👍

lib/chromic_pdf/pdf/browser/channel.ex
maltoe: https://github.com/bitcrowd/chromic_pdf/pull/84/files#diff-e39305e4b159a10e697177efaac4169b232d94600ea1cef173be63587083c968R51 

☝️ ☝️

lib/chromic_pdf/pdf/browser/session_pool.ex
maltoe: this is in fact [`Kernel.get_in/2`](https://hexdocs.pm/elixir/Kernel.html#get_in/2) which uses the Access behaviour and works with both keyword lists and maps.

PR #85, opened 2021-01-07
Simplify README and move code to example.

maltoe:
## TODOs

- [x] finish the loading of the styles into a module variable (or do this with `render` as well?)
- [x] add a b64-encoded image for demonstration


andreasknoepfle (approved):
Nice changes so far. I also love that the example is tested ❤️

maltoe:
@andreasknoepfle ty! i added some boilerplate for asset loading.. not sure how much of this we actually should put into the example, as it grows more and more complex. But I wanted to show somehow how to include static assets (images as data urls as you mentioned, but also stylesheets) in a PDF template, so maybe it has to be this complex? Plus it's of course only one possible way to do this out of a million, I'm worried this "example" would have too much influence on users? wdyt? maybe we can have a call on this at some point?

andreasknoepfle:
> But I wanted to show somehow how to include static assets (images as data urls as you mentioned, but also stylesheets) in a PDF template, so maybe it has to be this complex? 

Would it help breaking it into multiple examples? 

> Plus it's of course only one possible way to do this out of a million, I'm worried this "example" would have too much influence on users? 

I wouldn't worry about that too much. If someone gets the gist of it they will derive from this path soon enough :D. But often it is best to have those examples around as long as they all work and are not just abandoned code in some file (which I think is covered already by the testing part) 

> maybe we can have a call on this at some point?

Sure :)

examples/phoenix/phoenix_example.ex
```diff
@@ -0,0 +1,64 @@
...
+  @moduledoc false
+
+  def pdf do
+    quote do
+      use Phoenix.View, root: "examples/phoenix/templates"
+      import Phoenix.HTML
+      import PhoenixExample
+
+      def print_to_pdf(assigns, callback) do
+        [content: content(assigns)]
```
maltoe, line 11: page size?

examples/phoenix/assets/style.css
```diff
@@ -0,0 +1,3 @@
+img {
+  width: 400px;
+  height: 200px;
```
maltoe, line 3: closing }

PR #87, opened 2021-01-12
Add option to wait for specific attribute to be set before printing to PDF

### LINKED ISSUE #86: Wait for dynamic content to be ready before printing
We noticed that in some cases it is useful to delay the printing until dynamic content on the page is ready. This is admittedly rare but sometimes handy. The way we've approached it is to wait until a specific element has a defined attribute set.

Does the approach sound sane and would it make sense to include it in Chromic PDF? See the PR #87 for the approach.

jarimatti:
This PR adds an option to wait for a specific attribute value before printing the PDF. It's opt-in and documentation mentions that it's not recommended. The goal is to provide an escape hatch if static HTML template would be inconvenient for some reason.

Resolves #86


maltoe (changes requested):
Hey @jarimatti 

Thank you so much for the contribution! 💚 💛 💙  This is awesome!

Had a few minor comments (mostly on the docs) and a few questions. But overall I'm really impressed with this out-of-the-blue PR! Thank you!

Would you mind adding a line to `CHANGELOG.md` as well?

lib/chromic_pdf/pdf/protocols/print_to_pdf.ex
```diff
@@ -33,6 +33,31 @@ defmodule ChromicPDF.PrintToPDF do
       end
 
       await_notification(:frame_stopped_loading, "Page.frameStoppedLoading", ["frameId"], [])
+
+      if_option :wait_for do
```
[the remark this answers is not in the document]
jarimatti: Good question! Didn't think about that but I have a hunch that it does work. I'll verify it, thanks.

lib/chromic_pdf/supervisor.ex
```diff
@@ -158,18 +158,28 @@ defmodule ChromicPDF.Supervisor do
 
       @type telemetry_metadata_option :: {:telemetry_metadata, map()}
 
+      @type wait_for_option ::
+              {:wait_for,
+               %{
+                 required(:selector) => String.t(),
+                 required(:attribute) => String.t(),
+                 optional(:value) => String.t()
```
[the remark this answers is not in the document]
jarimatti: Good catch, that is correct and is a bug in typespec.

Initially I was just following for when the attribute is created without value: I'll fix the type and set the value mandatory.

lib/chromic_pdf/supervisor.ex
```diff
@@ -393,6 +403,24 @@ defmodule ChromicPDF.Supervisor do
 
       See [`print_header_footer_template.html`](https://cs.chromium.org/chromium/src/components/printing/resources/print_header_footer_template_page.html)
       from the Chromium sources to see how these values are interpreted.
+
+      ### Wait for dynamic content
+
+      You can wait for an attribute value to be set in a given element to enable dynamic content be ready.
+      This is inteded for special circumstances: it is recommended that the HTML is as static as possible.
```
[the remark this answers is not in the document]
jarimatti: Yeah, will remove this piece. In general I am a fan of "static HTML, no network" approach for this.

test/integration/fixtures/test_dynamic.html
```diff
@@ -0,0 +1,15 @@
+<html>
+
+<body>
+  <p id="print-ready" ready-to-print="false"></p>
```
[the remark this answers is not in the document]
jarimatti: No specific reason. The first version was in fact just following that the attribute is set, later I thought that perhaps a value could be useful. I'm happy with the idiomatic way, just following when the attribute appears. It would be more user-friendly, because there is no risk of typos in the value.

I can refactor this to just follow the attribute, would that be good approach?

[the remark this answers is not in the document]
maltoe: 🤔  I can't think of a use-case right now where the value would be truly needed, and as far as I can tell it's currently also not optional, so I would say yes, let's go without it.

test/integration/fixtures/test_dynamic.html
```diff
@@ -0,0 +1,15 @@
...
+
+  <script>
+    setTimeout(() => {
+      const element = document.getElementById('print-ready');
+      element.innerText = "Dynamic content from Javascript";
+      element.setAttribute('ready-to-print', true);
+    }, 1000);
```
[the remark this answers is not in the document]
jarimatti: I had a tingling thought about that and yes, it is a risk: definitely worth mentioning in the docs.

It could be mitigated by getting the attributes of the element first, just to verify that we can continue if the attribute is there. It would not hurt to add that to the protocol. AFAIK the developer protocol only sends events on elements that the client is aware.

A test could have static HTML page with the attribute set and it should pass.

test/integration/fixtures/test_dynamic.html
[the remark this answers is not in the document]
maltoe: another thought: Would it make sense to also give a really short sample of JavaScript in the documentation that triggers the attribute? I'm assuming you're using it with `window.onload` or so?

maybe sth. like

```
# the corresponding JS code in the markup
<script>
  window.onload = function() {
    // dynamic content loading...
    document.getElementById('print-ready').setAttribute('ready-to-print', 'true');
  };
</script>
```

test/integration/fixtures/test_dynamic.html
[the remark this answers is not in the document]
maltoe: 🤔  suggestion: focus on getting this PR ready first (maybe mention the race condition with a one-liner in the docs), and find a solution afterwards? Adding runtime conditionals to the protocol will be a bit messy, I think.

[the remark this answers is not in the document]
jarimatti: Thinking further I don't think my mitigation idea would work because of the missing "runtime conditionals" you mention: there is no way to branch in the protocol after checking the value.

lib/chromic_pdf/supervisor.ex
```diff
@@ -158,18 +158,28 @@ defmodule ChromicPDF.Supervisor do
 
       @type telemetry_metadata_option :: {:telemetry_metadata, map()}
 
+      @type wait_for_option ::
+              {:wait_for,
+               %{
+                 required(:selector) => String.t(),
```
maltoe, line 164: tiny last question: `String.t()` and `binary()` are equivalent AFAIK, but `String.t()` has the notion of "should always be UTF-8-encoded" -> are HTML attributes and selectors utf8 encoded?

test/integration/fixtures/test_dynamic.html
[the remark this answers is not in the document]
jarimatti: Yeah, I think that a documentation example would be good addition to the docs here. That's basically the gist of it: one could also follow the fonts being loaded etc.

lib/chromic_pdf/supervisor.ex
jarimatti: Good question, not that tiny... (be careful what you ask for) 😄  Here I used `String.t()` to tell that this here is a string, not arbitrary binary data.

Attribute names can contain quite a range of characters: https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

The selector API in JS uses strings but I haven't been able to confirm what exactly is allowed.

I added an integration test with ISO-8859 (latin-1) encoded HTML file and it does work with character `ö` in the attribute name and selector in both HTML and Elixir source.

jarimatti:
OK, I believe I've now addressed all the changes and this is ready for review.

lib/chromic_pdf/supervisor.ex
maltoe: hmm, given the specification doesn't say anything about character encoding, I think I'd prefer `binary()` here. Sorry to be so picky 😄

lib/chromic_pdf/supervisor.ex
```diff
@@ -393,6 +401,78 @@ defmodule ChromicPDF.Supervisor do
...
+
+      If your HTML fetches dynamic content via JavaScript, you can declare the content as "ready" by setting an attribute on a DOM element. You can wait for this attribute with the `wait_for` option.
+
+      The following example waits for the DOM element with ID `#element-id` to receive the attribute `ready-to-print` with value `true`.
+      before printing the page:
+
+          wait_for = %{
+            selector: "#element-id",
+            attribute: "ready-to-print",
+            value: "true"
```
maltoe, line 415:
```suggestion
```

Also, I suggest aligning the selector here with the example HTML below, and maybe make it `#dynamic-content`?

lib/chromic_pdf/supervisor.ex
```diff
@@ -393,6 +401,78 @@ defmodule ChromicPDF.Supervisor do
...
+            }
+          ]
+
+          ChromicPDF.print_to_pdf({:url, "http:///example.net"}, params)
+
+      ```html
+      <!-- Example element that gets dynamic content -->
+      <div id="dynamic"></div>
+      <!-- Marker element that gets the attribute. Can be same as the dynamic element. -->
+      <div id="print-ready"></div>
```
maltoe, line 439: Since it can be the same, please make it the same. Saves us 2 lines (+1 line in `loadDynamicContent`) and IMO this information doesn't add much value - in case users for whatever reason want to make this `print-ready` div "standalone", it is trivial to try out. But IMO the "default" way would be to just use the same div (or any div in your markup, really, as it doesn't matter...)

lib/chromic_pdf/supervisor.ex
```diff
@@ -393,6 +401,78 @@ defmodule ChromicPDF.Supervisor do
...
+          // Call any JS libraries/frameworks to set the dynamic content.
+          const dynamic_element = document.getElementById('dynamic');
+          dynamic_element.innerText = 'This is filled by JS';
+
+          // When ready, set the attribute on the element you're watching:
+          const element = document.getElementById('print-ready');
+          element.setAttribute('ready-to-print', '');
+        }
+
+        loadDynamicContent();
```
maltoe, line 452: otherwise nice example 👍  thank you! Also nice that you added some structure to the docs with the subheaders.

lib/chromic_pdf/supervisor.ex
```diff
@@ -393,6 +401,78 @@ defmodule ChromicPDF.Supervisor do
...
+      <script>
+        window.onload = () => {
+          // Simulate slow content load with 500 ms timeout.
+          setTimeout(() => {
+            const element = document.getElementById('print-ready');
+            element.innerText = "Dynamic content from Javascript";
+            element.setAttribute('ready-to-print', '');
+          }, 500);
+        }
+      </script>
```
maltoe, line 474: 🤔 

ok for now, I guess. I think I'd like to find a solution for this issue at some point (in other words, runtime conditionals)... I feel this caveat - which likely would go unnoticed by the majority of users - gets too much room in the docs now, and the solution is also quite hack-ish.

Nevertheless, let's leave it as-is for now.

test/integration/fixtures/test_dynamic.html
```diff
@@ -0,0 +1,26 @@
...
+
+  <script>
+    function loadDynamicContent() {
+      // Call any JS libraries/frameworks to set the dynamic content.
+      const dynamic_element = document.getElementById('dynamic');
+      dynamic_element.innerText = "Dynamic content from Javascript";
+
+      // When ready, set the attribute:
+      const element = document.getElementById('print-readyö');
+      element.setAttribute('ready-to-printö', '');
```
maltoe, line 19:
```suggestion
      const element = document.getElementById('print-ready');
      element.setAttribute('ready-to-print', '');
```

test/integration/pdf_generation_test.exs
```diff
@@ -131,6 +137,53 @@ defmodule ChromicPDF.PDFGenerationTest do
         path -> refute File.exists?(path)
       end
     end
+
+    # The wait_for tests intentionally use umlauts in selectors and attributes.
+    @tag :pdftotext
+    test "it waits until defined selectors have given attribute when printing from `:url`" do
+      params = [
+        wait_for: %{selector: "#print-readyö", attribute: "ready-to-printö"}
```
maltoe, line 145:
```suggestion
        wait_for: %{selector: "#print-ready", attribute: "ready-to-print"}
```

test/integration/pdf_generation_test.exs
```diff
@@ -131,6 +137,53 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+
+      print_to_pdf({:url, "file://#{@test_dynamic_latin1_html}"}, params, fn text ->
+        assert String.contains?(text, "Dynamic content from Javascript")
+      end)
+    end
+
+    @tag :pdftotext
+    test "it waits until defined selectors have given attribute when printing from `:html`" do
+      params = [
+        wait_for: %{selector: "#print-readyö", attribute: "ready-to-printö"}
```
maltoe, line 167:
```suggestion
        wait_for: %{selector: "#print-ready", attribute: "ready-to-print"}
```

test/integration/pdf_generation_test.exs
```diff
@@ -131,6 +137,53 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+      ]
+
+      print_to_pdf({:html, File.read!(@test_dynamic_html)}, params, fn text ->
+        assert String.contains?(text, "Dynamic content from Javascript")
+      end)
```
maltoe, line 172: Thank you for adding this test! 💚

test/integration/pdf_generation_test.exs
```diff
@@ -131,6 +137,53 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+        wait_for: %{selector: "#print-readyö", attribute: "ready-to-printö"}
+      ]
+
+      assert_raise(
+        RuntimeError,
+        fn ->
+          ChromicPDF.print_to_pdf({:url, "file://#{@test_dynamic_pre_existing_html}"}, params)
+        end
+      )
+    end
```
maltoe, line 186: For the others (`test_dynamic_latin1_html` and `test_dynamic_pre_existing_html`), could you please remove them again? 😄  Thanks for writing them and finding out about the ISO 8859 charset, but in the end I prefer not having to maintain them over the additional test coverage. Especially since neither of them really covers any functionality (= code) in ChromicPDF: One tests a feature of the HTML spec, the other actually tests a limitation in the library.

Again, hope you don't mind the pickyness.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Added
 
+- Option `wait_for` to wait for dynamic content using element attribute.
```
maltoe, line 5:
```suggestion
- Added option `wait_for` to wait for DOM element attribute to be set dynamically. (@jarimatti)
```

for the fame! 🥇

maltoe (changes requested):
👍  Thanks! Had another look and a few requests still, but overall looks really good. Again thanks for the contribution.

lib/chromic_pdf/supervisor.ex
jarimatti: Agreed with this being less than stellar hack. Ideally there would be runtime conditionals but it seems to be slightly bigger thing to tackle. I could remove the example from the caveat completely: we can probably assume that people using this library know their way around JS and HTML?

lib/chromic_pdf/supervisor.ex
jarimatti: I knew I missed something! Sigh...

lib/chromic_pdf/supervisor.ex
maltoe: as you prefer, yes. I think I'll take a look into runtime conditionals when I find some time and would then hopefully remove this entire section.

lib/chromic_pdf/supervisor.ex
maltoe: I think we made a good call on removing the `value`, by the way. Makes the feature a lot simpler to grasp & use.

test/integration/pdf_generation_test.exs
jarimatti: No worries regarding the pickyness, appreciate your diligence. It's more or less expected when hopping into unfamiliar code without initial discussions with the maintainers.

I'll remove these additions, no need to take the burden with them. I should have time this evening to make the changes.

maltoe (approved):
💯  thanks again!

PR #94, opened 2021-01-15
[92] Option to ignore certificate errors

### LINKED ISSUE #92: Option to ignore certificate errors
Hi again!

In puppeteer you can pass `ignoreHTTPSErrors: true` to the launch config which sends the command
'Security.setIgnoreCertificateErrors', {ignore: true}.

It would be great to have this option in chromic_pdf too.

maltoe:
Fixes #92.


test/integration/pdf_generation_test.exs
```diff
@@ -165,7 +165,7 @@ defmodule ChromicPDF.PDFGenerationTest do
 
     @tag :pdftotext
     test "it does not print PDFs from https:// URLs when given the offline: true parameter" do
-      assert_raise RuntimeError, ~r/net::ERR_INTERNET_DISCONNECTED/, fn ->
+      assert_raise ChromicPDF.ChromeError, ~r/net::ERR_INTERNET_DISCONNECTED/, fn ->
```
andreasknoepfle, line 168: would it maybe make sense to also assert that the error message contains the explanatory text?

test/integration/support/test_server.ex
```diff
@@ -0,0 +1,34 @@
...
+  @moduledoc false
+
+  use Plug.Router
+
+  @port 44_285
+
+  @cowboy_opts [
+    http: [port: @port],
+    https: [
+      port: @port,
```
andreasknoepfle, line 11: feels a bit weird how `http` and `https` apparently run on the same port. How can this even work?

andreasknoepfle (approved):
Awesome works!

test/integration/support/test_server.ex
maltoe: They never run at the same time :)

can make it different ports though, if that makes it more clear

andreasknoepfle: Can the ports be somehow auto assigned to a free one, like sockets to when assigning `0` as a port? Would be great.

test/integration/pdf_generation_test.exs
maltoe: meh... could do, but there are ton's of newlines in there. I think I'm ok with asserting that it is a `ChromeError` :)

andreasknoepfle: I also thought about maybe the first sentence or so, just to make sure a comment is shown and not only the original error. But as you like :)

maltoe: Ok, done :+1:

TIL: `~r/foo.*bar/s` can match `"foo\n\n\nbar"`

test/integration/support/test_server.ex
maltoe: 🤔  good idea. It is possible, but I don't know how to get ahold of the used port unfortunately. plug_cowboy doesn't have any functions for it. So, yes, nice to have, but IMO not necessary to put a lot of time into.

PR #95, opened 2021-01-19
[91] Custom options to chromium

### LINKED ISSUE #91: Custom options to chromium
Hi again!

We need to pass a custom option to chromium on startup, in this case we need `"--font-render-hinting=none"`. I don't see an immediate way to do this.

See https://github.com/puppeteer/puppeteer/issues/2410 for why we need font-render-hinting=none. In any case, it would be useful to be able to pass custom options.

maltoe:
Fixes #91.


andreasknoepfle (approved):
Neat 👍

PR #97, opened 2021-01-20
Count targets in Chrome instead of OS pids.

maltoe:


andreasknoepfle:
So what does this do exactly? I thought this is just a flaky test fix :D

maltoe:
> So what does this do exactly? I thought this is just a flaky test fix :D

:) It does, among other things. Counting the OS processes of Chrome and comparing their pids never was a reliable solution (for instance, if in development you closed your Chrome browser while running these tests, they would fail as well), so I figured fetching the IDs of targets from Chrome itself was a much better solution. However, that's not what fixed the flaky test, it needed a few more wait statements here and there. And I'm not even certain it is 100% fixed on CI 😄 we'll see.

Can explain you the details in a call, if needed.

test/integration/session_restart_test.exs
```diff
@@ -9,25 +10,24 @@ defmodule ChromicPDF.SessionRestartTest do
     end
 
     test "session restart spawns a new session process" do
-      pids_before = chrome_renderer_pids()
+      targets_before = GetTargets.baseline()
```
maltoe, line 13: wait until 1 instead of 100ms?

maltoe:
@andreasknoepfle updated, waiting until `length(targets)` is pool size now (instead of 100ms baseline)

andreasknoepfle (approved):

PR #102, opened 2021-02-05
[101] Get number of schedulers at runtime

maltoe:


andreasknoepfle (approved):
Looks awesome afaics 👍 💯

PR #106, opened 2021-02-23
[105] Await_response not failing when payload does not contain key

### LINKED ISSUE #105: await_response not failing when payload does not contain key
See stacktrace at #104 

* `await_response` and `await_notification` both call `extract_from_payload`
* `extract_from_payload` doesn't error when it can't find the desired key in the payload, but instead sets it to `nil` in the state
* When the value is to be used again in a follow-up call, `get_in_state!` errors out when the value is `nil`.
* FIX: error out when `await_response` or `await_notification` can't satisfy their payload extractions

maltoe:
Closes #105.

### DIFF
--- a/lib/chromic_pdf/pdf/protocol_macros.ex
+++ b/lib/chromic_pdf/pdf/protocol_macros.ex
@@ -123,7 +123,7 @@ defmodule ChromicPDF.ProtocolMacros do
   end
 
   defp fetch_param_for_call(state, {name, key_path}) do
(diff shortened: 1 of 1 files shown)


andreasknoepfle (approved):
Seems legit :)

andreasknoepfle:
Maybe this also want a changelog 🤷 ?

maltoe:
> Maybe this also wants a changelog 🤷 ?

IMO it's not user-facing, so I'd not write one

PR #111, opened 2021-02-26
[107] Compile time debug option to log calls and Chrome messages

### LINKED ISSUE #107: Compile time debug option to log calls and Chrome messages
From @jarimatti in #104 

> Would be nice if we could log the DevTools protocol messages for diagnostics via some means, e.g. config flag or some other setting: the messages should shed some light into this. Not sure if Chrome can do that natively?

Agreed. I'm also constantly going back to the [`Channel`](https://github.com/bitcrowd/chromic_pdf/blob/066da34f8fb24e04a3ad201d5023cd5d515b4b67/lib/chromic_pdf/pdf/browser/channel.ex#L67) and [`Connection`](https://github.com/bitcrowd/chromic_pdf/blob/066da34f8fb24e04a3ad201d5023cd5d515b4b67/lib/chromic_pdf/pdf/connection/dispatcher.ex#L18) to put in `IO.inspect`s. I suggest adding a compile time switch that conditionally compiles in some debug code, either `Logger.debug` or just direct `IO.inspect`s.

maltoe:
Closes #107

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,5 +1,8 @@
 import Config
 
+# Set this to true to see protocol messages.
+config :chromic_pdf, debug_protocol: false
+
 if Mix.env() == :test do
   config :chromic_pdf, chrome: ChromicPDF.ChromeMock
 end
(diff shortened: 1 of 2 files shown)


andreasknoepfle (approved):
Clever 👍

PR #113, opened 2021-03-11
[#112] Replace `wait_for` with script evaluation

### LINKED ISSUE #112: Replace `wait_for` with script evaluation
Hi @jarimatti ,

posting this here so we can discuss it and you feel involved and not replaced (because replacing is what I intend to do 😄 ):

## on fixing the race condition in `wait_for`

I looked into the `wait_for` option a bit more and found out that in its current implementation it's definitely a bit cumbersome to use, and fixing it is unfortunately non-trivial. 

* As we discussed before, if the element already has the attribute set before we send the `querySelector` call, we won't get any `attributeModified` notification and run into a timeout. In fact, the "Example HTML" we have in the [docs](https://hexdocs.pm/chromic_pdf/ChromicPDF.html#print_to_pdf/2) has this error if used as-is (i.e. without any `setTimeout` or JS lib doing some work in between).
* We could fix this two ways:
  - "polling": Repeatedly ask Chrome to [`DOM.getAttributes`](https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getAttributes) instead of waiting for `attributeModified` -> this should work, but feels a little clunky.
  - Reading the `DOM.setChildNode` notifications we get right after `DOM.getDocument` and filter out the relevant node and its attributes. This would be my preferred approach, but turned out to be rather tricky unfortunately
    - the `DOM.setChildNode` messages come after `DOM.getDocument`, i.e. theoretically before we get the result of `DOM.querySelector` -> hence at this point in time we might not know the `nodeId` of the element in question. We could instead match the node based on the `id` attribute here.
    - `DOM.setChildNode` notifications traverse the tree, so technically we will get between one and many of them, and would need to read them all. That's another thing that the `Protocol` machine can't do right now.

In summary: Either way outlined above, in combination with some kind of `if_state` runtime conditional, is a whole lot of complexity we would introduce only to make sure the `wait_for` logic is race-free, i.e. can deal with elements which have the attribute already set. Therefore, even if I like the API interface quite a lot (as it covers a lot and doesn't need help from the client-side), **IMO** it is not worth the added complexity in the code. So, at this point I discarded the entire idea and looked for alternatives.

## Evaluating a custom script instead

Actually I had planned to support this devtools call even before you added `wait_for`, but didn't move on with it as I personally had no use for it at the time: [`Runtime.evaluate`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-evaluate). It might be a bit of a sledgehammer solution to replace `wait_for`, but of course it has potential to solve other use-cases as well. Looking into it this morning, I wanted to specifically see what it takes to mimick `wait_for` though, and here it is:

```elixir
# protocols/print_to_pdf.ex
if_option :evaluate do
  call(:evaluate, "Runtime.evaluate", [{"expression", [:evaluate, :expression]}], %{awaitPromise: true})
  await_response(:evaluated, [])
end
```

```
# client call
@wait_for_js """
async function waitForAttribute(selector, attribute) {
  while (!document.getElementById(selector).hasAttribute(attribute) {
    await new Promise(resolve => requestAnimationFrame(resolve));
  }
};

waitForAttribute('testdiv', 'ready-to-print');
"""

ChromicPDF.print_to_pdf(..., evaluate: %{expression: @wait_for_js})
```

Of course I didn't come up with this script myself: [stackoverflow link](https://stackoverflow.com/a/53269990). In my test setup, this worked for both cases; when I had the attribute already set on the element as well as with `setTimeout(... setAttribute..., 1000)`.

In my opinion, this is quite neat. What do you think? Would you be ok with replacing the `wait_for` option with this instead?One could theoretically also emulate the `wait_for` behaviour with this script underneath, i.e. put this script into ChromicPDF and make the `wait_for` option set an `evaluate` option instead. But I'm not yet convinced maintaining this in code has much added benefit over just mentioning it in the docs.

maltoe:
Fixes #112.


lib/chromic_pdf/api.ex
```diff
@@ -35,6 +35,7 @@ defmodule ChromicPDF.API do
     opts =
       opts
       |> put_source(source)
+      |> rewrite_wait_for()
```
andreasknoepfle, line 38: `rewrite_wait_for` seems like a weird naming. Sounds like an issue title rather than a function name. 🤷 
But on the other hand I can't come up with a better one. `replace_wait_for_with_evaluate` or simply `handle_wait_for`. But maybe it is also ok.

lib/chromic_pdf/api.ex
```diff
@@ -60,6 +61,32 @@ defmodule ChromicPDF.API do
...
+
+    Keyword.put(opts, :evaluate, %{
+      expression: """
+      const waitForAttribute = async (selector, attribute) => {
+        while (!document.querySelector(selector).hasAttribute(attribute)) {
+          await new Promise(resolve => requestAnimationFrame(resolve));
+        }
+      };
+
+      waitForAttribute('#{selector}', '#{attribute}');
```
andreasknoepfle, line 85: Would it make sense to put such half js things into a `js.eex` template, so editors can syntax highlight them better?

The module is already quite big, so maybe putting some of it into a separate file would be nice :)

lib/chromic_pdf/api.ex
```diff
@@ -60,6 +61,32 @@ defmodule ChromicPDF.API do
     |> Keyword.put_new(source_type, source)
   end
 
+  defp rewrite_wait_for(opts) do
+    opts
+    |> Keyword.pop(:wait_for)
+    |> do_rewrite_wait_for()
```
andreasknoepfle, line 67: I came to the habit on piping these into a `case`, so you do not need the `do_` version.

lib/chromic_pdf/pdf/protocols/print_to_pdf.ex
```diff
@@ -35,28 +35,20 @@ defmodule ChromicPDF.PrintToPDF do
...
-        [
-          {["nodeId"], "selectorNodeId"},
-          {["name"], [:wait_for, :attribute]}
-        ],
-        []
-      )
+          %{"exception" => %{"description" => error}} ->
+            {:error, "[evaluate] #{error}"}
+        end
+      end
```
andreasknoepfle, line 51: 💯

lib/chromic_pdf/supervisor.ex
```diff
@@ -451,50 +454,37 @@ defmodule ChromicPDF.Supervisor do
...
-      #### Caveats
+          wait_for = %{
+            selector: "#my-element",
+            attribute: "ready-to-print"
+          }
 
-      If the attribute already exists on the element the PDF, generation will fail with a timeout. This could
-      happen if the attribute is set too quickly, before ChromicPDF has the chance to observe the element.
-      As a workaround you can use `setTimeout` with a small delay.
+          ChromicPDF.print_to_pdf({:url, "http:///example.net"}, wait_for: wait_for)
```
andreasknoepfle, line 487: Nice API :)

lib/chromic_pdf/supervisor.ex
```diff
@@ -451,50 +454,37 @@ defmodule ChromicPDF.Supervisor do
...
-      The following example waits for the DOM element with ID `#my-element` to receive the attribute `ready-to-print`
-      before printing the page:
+      In case your print source is generated by client-side scripts, for instance to render
+      graphics or load additional resources, you can trigger these by evaluating a JavaScript
+      expression before the PDF is printed.
 
-          wait_for = %{
-            selector: "#my-element",
-            attribute: "ready-to-print"
+          evaluate: %{
```
andreasknoepfle, line 465:
```suggestion
          evaluate = %{
```

lib/chromic_pdf/supervisor.ex
```diff
@@ -451,50 +454,37 @@ defmodule ChromicPDF.Supervisor do
...
+            expression: """
+            document.querySelector('body').innerHTML = 'hello world';
+            """
```
andreasknoepfle, line 468: maybe it could even be nice to have some kind of helper to load `js.eex` templates, that we use ourselves :)

andreasknoepfle (approved):
Awesome 🚀 , some thoughts, some nitpicks and a documentation issue :) Otherwise looks marvelous as always

lib/chromic_pdf/api.ex
maltoe: i totally agree on the "too big" part, only having troubles identifying a suitable name / concept for a different module. Something along the lines of `BrowserOptions` or so (to keep everything in `chrome_export`), but it whenever I move these functions somewhere else they feel really out of place by themselves. Weird functionality on some obscure `opts` dict...

lib/chromic_pdf/api.ex
maltoe: :) 👍  makes sense, but I kind of like the `do_` version here as otherwise I have so much indentation

lib/chromic_pdf/api.ex
maltoe: I moved it into an eex template but left it in the same file 😄  Probably not what you wanted, but closer to it. If you can figure out a good name / concept to extract from the `API` module, pls go ahead

PR #114, opened 2021-03-13
Shared protocol fragments.

maltoe:
* Add new `include_protocol` macro to the `ProtocolMacros` which builds
  a shared protocol and merges its steps into the current protocol.
* Extract `Navigate` and `Blank` protocols from `PrintToPDF` and use
  them in `CaptureScreenshot` (gains all navigational options from
  `print_to_pdf/2`!) and `SpawnSession`.

### DIFF
--- a/lib/chromic_pdf/supervisor.ex
+++ b/lib/chromic_pdf/supervisor.ex
@@ -179,14 +179,16 @@ defmodule ChromicPDF.Supervisor do
                  required(:attribute) => binary()
                }}
 
-      @type pdf_option ::
-              {:print_to_pdf, map()}
-              | {:set_cookie, map()}
-              | info_option()
-              | output_option()
-              | telemetry_metadata_option()
+      @type navigate_option ::
+              {:set_cookie, map()}
               | evaluate_option()
               | wait_for_option()
+
+      @type pdf_option ::
+              {:print_to_pdf, map()}
+              | navigate_option()
+              | output_option()
+              | telemetry_metadata_option()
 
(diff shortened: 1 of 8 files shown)


lib/chromic_pdf/pdf/protocols/blank.ex
```diff
@@ -0,0 +1,24 @@
+defmodule ChromicPDF.Blank do
```
andreasknoepfle, line 1:
```suggestion
defmodule ChromicPDF.Reset do
```
like in `CSS` :P

or maybe `ResetTarget` ?

lib/chromic_pdf/supervisor.ex
```diff
@@ -179,14 +179,16 @@ defmodule ChromicPDF.Supervisor do
                  required(:attribute) => binary()
                }}
 
+      @type navigate_option ::
```
andreasknoepfle, line 182: I just realized wouldn't they be all plural btw?  `options`

andreasknoepfle (approved):
Nice 💯 I love the refactoring brings new features vibe here 🏅

lib/chromic_pdf/supervisor.ex
maltoe: Nah, it's read like this: "A single `navigate_option` is either this or that or that" -> Then where these are used in the function specs like

```
@spec print_to_pdf(..., [pdf_option()])
```

ie. `print_to_pdf` receives a list of `pdf_option` thingys

PR #116, opened 2021-03-17
Divide up API module and improve JS errors.

maltoe:


lib/chromic_pdf/api/pdf_options.ex
```diff
@@ -0,0 +1,117 @@
+defmodule ChromicPDF.PDFOptions do
+  @moduledoc false
+
+  import ChromicPDF.Utils
+  require EEx
+  alias ChromicPDF.ChromeError
+
+  def prepare_export_options(source, opts) do
```
andreasknoepfle, line 8: If the modules would be aliased instead of imported, the public methods in here might be ok with less verbose names. E.g. `ChromeOptions.build` 🤷 
Also where is `feed_chrome_data_into_output` even used? Didn't find it in the PR. Might because of the import.

lib/chromic_pdf/api/chrome_error.ex
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Exception:
+
+    #{indent(description)}
+
+    Evaluated expression:
+
+    #{indent(expression, line_number)}
+    """
+  end
```
andreasknoepfle, line 65: 👏

lib/chromic_pdf/api/pdfa_options.ex
```diff
@@ -0,0 +1,22 @@
...
+
+      nil ->
+        data =
+          pdfa_path
+          |> File.read!()
+          |> Base.encode64()
+
+        {:ok, data}
+    end
+  end
```
andreasknoepfle, line 21: Reading this I was close to thinking if one could refactor the option handling to feel like a middleware stack, where each middleware would be handled on the way there and back to process specific options. 
Like this one could separate all this concerns into very small modules. 
But just a thought on the side, no action needed :)

test/integration/pdf_generation_test.exs
```diff
@@ -165,10 +165,37 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+
+            function foo() {
+      !!!     throw new Error(\"boom\");
+            }
+            foo();
+
+      """
 
-      assert_raise ChromicPDF.ChromeError, ~r/SyntaxError: Unexpected token '}'/, fn ->
+      assert_raise ChromicPDF.ChromeError, expected_msg, fn ->
```
andreasknoepfle, line 198: 👍

andreasknoepfle (approved):
Awesome change :) 

I feel like the improving of JS error smells like changelog entry.

lib/chromic_pdf/api/pdf_options.ex
maltoe: See that was my whole point about being unsure how to split it up: If it was clearly only about the "ChromeOptions" as an entity, I could also imagine a `ChromeOptions.build`... but the `feed_*` functions aren't concerned so much with the options for chrome anymore, but rather deal with the `output: ` option only. In the end all of these functions are just helpers, completely meaningless without the context of the `API` module. So in the end: Idk, code has been separated, I don't feel like any more renames provide any benefit. Suggestions welcome of course.

maltoe:
> I feel like the improving of JS error smells like changelog entry.

:) true, but since the `evaluate` option was just added in this release, it would be weird to already have a "Changed" entry for it, wouldn't it?

lib/chromic_pdf/api.ex
```diff
@@ -1,9 +1,17 @@
 defmodule ChromicPDF.API do
   @moduledoc false
 
-  import ChromicPDF.Utils
-  require EEx
-  alias ChromicPDF.{Browser, CaptureScreenshot, ChromeError, GhostscriptPool, PrintToPDF}
+  import ChromicPDF.{Telemetry, Utils}
+
+  # credo:disable-for-next-line Credo.Check.Readability.AliasOrder
```
maltoe, line 6: credo is weird:

```
┃ [R] ↘ The alias `ChromicPDF` is not alphabetically ordered among its group.
┃       lib/chromic_pdf/api.ex:7:9 #(ChromicPDF.API)
```

seems like a bug

PR #122, opened 2021-04-16
Allow specification of custom chrome executeable

resterle:


test/unit/chromic_pdf/pdf/connection_test.exs
```diff
@@ -29,6 +29,12 @@ defmodule ChromicPDF.ConnectionTest do
       expect(ChromeMock, :spawn, fn ^opts -> {:ok, @port} end)
       assert init({self(), opts}) == {:ok, new_state()}
     end
+
+    test "it spawns custom Chrome executeable and initializes its state" do
+      opts = [discard_stderr: false, no_sandbox: true, chrome_executeable: "/cutom/chrome", chrome_args: "--foo"]
```
maltoe, line 34: Maybe let's just add this `chrome_executable` option to the test above. These unit tests are a bit pointless in general, they mostly assert that the args are passed to `ChromeImpl` (i.e. the `Keyword.take` call).

maltoe (changes requested):
Hey @resterle 👋 

Thanks for the contribution! Code looks good overall, but I have a few requests:

1. Please let's be consistent in `executeable` vs. `executable` -> `System.find_executable` doesn't have an `e`, so let's also not have one.
2. Please add a short description of the new option to the docs [here](https://github.com/bitcrowd/chromic_pdf/blob/main/lib/chromic_pdf.ex#L199-L205). You can just amend this section, e.g. "Custom executable path and command line switch"
3. Please add the option to the typespecs [here](https://github.com/bitcrowd/chromic_pdf/blob/main/lib/chromic_pdf/supervisor.ex#L215)
4. Please add a note to the [CHANGELOG](https://github.com/bitcrowd/chromic_pdf/blob/main/CHANGELOG.md) (add a new `## Added` section to the unreleased logs)

test/unit/chromic_pdf/pdf/connection_test.exs
resterle: Hi,

thank you for your quick response 👍
I implemented the requested changes.

Thanks and have a great weekend.

maltoe (approved):
👍  code looks good now, thanks again! 💯 

The linter wants a `mix format` though to be fully happy :)

resterle:
I think the linter is happy now^^

maltoe:
@resterle do you need a new release or can that wait for a bit?

resterle:
> @resterle do you need a new release or can that wait for a bit?

A new release would be great if it does not make to much work.

maltoe:
released v1.1.0

resterle:
Thank you!

PR #126, opened 2021-08-23
Add Chromium path on macOS

shamanime:
This adds the Chromium path for macOS users.

Edit: reference is https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/eloston-chromium.rb that does not add the binary to $PATH.


Schultzer:
FYI: This path are common for people installing chromium via HomeBrew:

https://formulae.brew.sh/cask/eloston-chromium
https://formulae.brew.sh/cask/chromium
https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/freesmug-chromium.rb

Schultzer (approved):

PR #139, opened 2022-01-22
[#136] Support Stream

### LINKED ISSUE #136: Support Stream
Thanks for the awesome job
Really useful lib

I'm bringing here what we discussed on the forum:

Sometimes I need to print very large pdf files and the memory consumption jumps too high

It would be nice if we could support [transferMode as ReturnAsStream](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF)

(I've tried to look at the code but, because of my lack of experience in elixir, I could not find an easy way to implement IO.read/IO.close in the current macro schema)

maltoe:
Closes #136

This patch adds `print_to_pdf_as_stream`, a new PDF printing function
that makes use of the `transferMode: "ReturnAsStream"` option.

Drive-by refactors:

* Simplify `SessionPool` and make it responsible only for session
  checkin/checkout and keeping track of the use count.
* Move the rest of the functionality (spawn protocol, etc.) to `Channel`
  with an API in `Browser`.
* Expose the `checkout_session` function to the API layer so it can use
  it for `print_to_pdf_as_stream` (as that needs to run multiple
  protocols on the same session).
* Dropped the `checkout_and_count` vs. `checkout` thing in favour of simply counting each session use.

## TODOs

- [ ] try it out, see if it does any good
- [ ] refactor `print_to_pdf_as_stream`, move the code somewhere else
- [ ] add tests for `print_to_pdf_as_stream`
- [ ] add an option validation to both functions (forbid `transferMode`)
- [ ] fix the fallout of the `checkout_and_count` removal
- [ ] add to changelog

### DIFF
--- a/lib/chromic_pdf/api.ex
+++ b/lib/chromic_pdf/api.ex
@@ -7,19 +7,23 @@ defmodule ChromicPDF.API do
   alias ChromicPDF.{
     Browser,
     CaptureScreenshot,
+    CloseStream,
     GhostscriptPool,
     PDFOptions,
     PDFAOptions,
-    PrintToPDF
+    PrintToPDF,
+    PrintToPDFAsStream,
+    ReadFromStream
   }
 
   @spec print_to_pdf(
           ChromicPDF.Supervisor.services(),
           ChromicPDF.source() | ChromicPDF.source_and_options(),
-          [ChromicPDF.pdf_option()]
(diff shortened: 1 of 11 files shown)


maltoe:
This doesn't have the desired effect. And is quite complicated.

Issue #129, opened 2021-09-22
Generate landscape PDF


cblavier:
Hi there,

I'm currently struggling at generating a PDF from an url in landscape mode.
I tried with the following options:
```
ChromicPDF.print_to_pdf(
  %{
    source: {:url, "http://acme.com/pdf"},
    opts: [
      print_to_pdf: %{landscape: true}
    ]
  }
)
```

or 

```
ChromicPDF.print_to_pdf(
  %{
    source: {:url, "http://acme.com/pdf"},
    opts: [
      print_to_pdf: %{paperWidth: 11, paperHeight: 8.5}
    ]
  }
)
```

None is working. Any idea?

andreasknoepfle:
Which version of chromic_pdf are you using? 
In the most recent 1.1.0 there is only `print_to_pdf/2` in the docs https://hexdocs.pm/chromic_pdf/1.1.0/ChromicPDF.html#print_to_pdf/2

```
ChromicPDF.print_to_pdf(
  {:url, "http://acme.com/pdf"},
  print_to_pdf: %{landscape: true}
 )
```

andreasknoepfle:
I just tried out your example and for me it does generate landscape layouts:

I use chrome on a mac with this version: 
```
Version 93.0.4577.82 (Official Build) (x86_64)
```

[landscape.pdf](https://github.com/bitcrowd/chromic_pdf/files/7210002/landscape.pdf)
[no-landscape.pdf](https://github.com/bitcrowd/chromic_pdf/files/7210003/no-landscape.pdf)

Could you give some more context on your environment (Chrome-Version, OS, etc ...) ?

cblavier:
I'm using Ubuntu 20.04.3 LTS with Google Chrome 94.0.4606.54

andreasknoepfle:
Hmm, I doubt that something is different on the chrome side. Could you build a minimal example project where landscape does not work (since I guess you have probably some more code around it), so we can reproduce this and help? 

I just tried as well with Chrome on a mac  version `94.0.4606.54 (Official Build) (x86_64)` and still can produce landscape pdfs 🤷  so it is a bit hard to reproduce your problem.

cblavier:
Thanks for your help, the landscape option was correctly set and propagated.
The issue was a CSS issue

maltoe:
Glad you could figure it out!

> The issue was a CSS issue

Anything we should put into the documentation?

cblavier:
[...]

PR #140, opened 2022-03-11
Fix landscape issue

williamthome:
I figured out [this issue](https://github.com/bitcrowd/chromic_pdf/issues/129).
The ```landscape: true``` option takes no effect for me. If this flag is included in ```print_to_pdf``` options, the paper doesn't rotate in my implementation.
See [this (maybe) related issue](https://github.com/puppeteer/puppeteer/issues/3834) on puppeteer.
Any suggestions?

Also updated the dependencies, but CI action fails, then I added the clean option to the CI config file, and all works as expected.


.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ jobs:
             - v2-deps-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
             - v2-deps-cache-{{ .Branch }}
             - v2-deps-cache
-      - run: mix deps.get
+      - run: mix do deps.clean --all, deps.get, deps.compile
```
maltoe, line 22: this doesn't make much sense I'm afraid

* `deps.clean --all` is going to *empty* the `deps` directory (the cache we've just restored in the line above)
* `deps.compile` is cool, but doesn't need to happen at this point. Dependencies are compiled as part of `mix compile` and cached in the "build cache"

please revert this

lib/chromic_pdf/api/chrome_error.ex
```diff
@@ -75,10 +75,9 @@ defmodule ChromicPDF.ChromeError do
     |> String.trim()
     |> String.split("\n")
     |> Enum.with_index()
-    |> Enum.map(fn
+    |> Enum.map_join("\n", fn
```
maltoe, line 78: 👍

lib/chromic_pdf/template.ex
```diff
@@ -225,7 +257,8 @@ defmodule ChromicPDF.Template do
   def styles(opts \\ []), do: do_styles(opts)
 
   defp do_styles(opts) do
-    {width, height} = get_paper_size(opts)
+    landscape = Keyword.get(opts, :landscape, false)
```
maltoe, line 260: instead of doing this twice, how about we extract `landscape` in `get_paper_string(opts)`?

lib/chromic_pdf/template.ex
```diff
@@ -244,13 +277,27 @@ defmodule ChromicPDF.Template do
...
-  defp get_paper_size(manual) when tuple_size(manual) == 2, do: manual
-  defp get_paper_size(name) when is_atom(name), do: Map.fetch!(@paper_sizes_in_inch, name)
+  defp get_paper_size(manual) when tuple_size(manual) === 2, do: manual
+
+  defp get_paper_size(name) when is_atom(name) do
+    @paper_sizes_in_inch
+    |> Map.get(name, @default_paper_size)
+  end
 
   defp get_paper_size(opts) when is_list(opts) do
```
maltoe, line 292: i.e. in here

lib/chromic_pdf/template.ex
```diff
@@ -43,12 +58,28 @@ defmodule ChromicPDF.Template do
...
-    us_letter: {8.5, 11.0}
+    a5: {5.8, 8.3},
+    a6: {4.1, 5.8},
+    a7: {2.9, 4.1},
+    a8: {2.0, 2.9},
+    a9: {1.5, 2.0},
+    a10: {1.0, 1.5},
+    us_letter: {8.5, 11.0},
+    legal: {8.5, 14.0},
+    tabloid: {11.0, 17.0}
```
maltoe, line 77: 💜

maltoe (changes requested):
Hey @williamthome 

Awesome work! Thank you!

IMO a bit strange that `landscape` isn't taking effect if paper size is given explicitly, but maybe that makes sense from Chrome's perspective. The changes you proposed seem like a good solution to me.

A couple requests though:

* Could you add a test for it (in `pdf_generation_test.exs`)?

`pdfinfo` shows two related fields, `Page size` and `Page rot`. I'd be interested in both of these with the inverted page size.

```
Page size:       594.96 x 840.96 pts (A4)
Page rot:        0
```

I'd suspect it'll show this for your "inverted page size":

```
Page size:       840.96 x 594.96 pts
Page rot:        0
```

iow it'll not say the page is rotated (because landscape was discarded) and it won't detect the A4 format anymore (because that is given in portrait). Not sure if this has any effects on PDF readers, but I guess it should be fine?

* Could you also add a changelog for this?
* Super happy that you decided to give the library some love overall, but could we do the dependency update in a separate PR? 🙏 

Again, thanks a lot for your contribution! 💜

.circleci/config.yml
williamthome: Hi @maltoe! The CI failed because of this.
Above the CI error:
```
#!/bin/bash -eo pipefail
mix test
** (Mix) Could not start application ranch: exited in: :ranch_app.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) function :ranch_app.start/2 is undefined (module :ranch_app is not available)
            (ranch 1.8.0) :ranch_app.start(:normal, [])
            (kernel 7.1) application_master.erl:277: :application_master.start_it_old/4

Exited with code exit status 1
CircleCI received exit code 1
```
Using `- run: mix do deps.clean --all, deps.get, deps.compile` the error is gone.

williamthome: I think the cache works anyways.
The first workflow takes 6m58s and the rest a max of 2m30s.
https://app.circleci.com/pipelines/github/williamthome/chromic_pdf?filter=all

williamthome:
The `Page rot` field of `pdfinfo` is always zero, even setting `landscape: true` at `print_to_pdf` options.
I forget to update the changelog. I'll take a look at this.

.circleci/config.yml
andreasknoepfle: You should have the same effect, by simply changing the cache version from v2 to v3 (which is usually how to clear the cache on circleci) 
I guess the cache has been build with an older erlang version I think, therefore we end up with this error every time. 
Your code will still use the cache for downloading, but not for compilation, which is still a huge portion of the time.

maltoe: > I think the cache works anyways. The first workflow takes 6m58s and the rest a max of 2m30s.

The speed-up you've seen is likely due to the other caches (compile cache, dialyzer, etc.), because each of these will have been rebuilt with the updated Erlang version and hence are now faster again. The deps cache is definitely useless if we delete the deps right after restoring them from cache ;)

maltoe: > changing the cache version from v2 to v3

👍  Though I guess now that we have rebuilt everything already we don't have to bump the cache number anymore. Just removing this change should be fine.

williamthome: Ok. 
I reverted the change at 4aa7afaa05ddd3ad2839ac30c00fec59bab42114.
Should mark this as resolved? Or how I can resolve this issue?

maltoe:
@williamthome we'll merge this tomorrow and fix the CI for you

thanks for the contribution 💜

andreasknoepfle:
Released in https://hexdocs.pm/chromic_pdf/1.2.0/readme.html

PR #147, opened 2022-03-17
Introduce new session_pool init_timeout option

dvic:


dvic:
On our CI (where things slow down quite a bit because of parallel tests) we noticed some failing builds from time to time. Upon further inspection it turned out that the session pool timeout option was ignored on startup. This PR fixes that by using the configured value also on startup.

dvic:
The test *session pool timeout can be configured and generates a nice error messages* is failing because the timeout now occurs during startup and not during the call to `print_to_pdf/1` I guess? I'm not sure how we can test this properly. Any ideas?

maltoe:
Hi @dvic 👋 

Thanks for looking into this!

Two ideas:

* You could rewrite the test to have a sensible timeout (say 100ms, not 1ms) and just print a really long document
* The "omission" you fixed here was actually intended. When writing this, I always considered a long pool init time (i.e. the spawn session protocol) to be a bug or at least some critical situation people would like to know about, whereas the "operation timeout" (as it is called in the docs) can certainly be higher due to "normal use". To account for that, we could add another option (say `session_pool: [init_timeout: ...]` or so) instead of using the existing one. That way you could even write tests for both.

While probably a bit more work, I'd prefer option 2. What do you think?

cheers,
malte

dvic:
> Hi @dvic 👋
> 
> Thanks for looking into this!
> 
> Two ideas:
> 
> * You could rewrite the test to have a sensible timeout (say 100ms, not 1ms) and just print a really long document. Not the prettiest solution, but it should be possible to find a timeout and a document length that make the test reliably run into the inner timeout.
> * The "omission" you fixed here was actually intended. When writing this, I always considered a long pool init time (i.e. the spawn session protocol) to be a bug or at least some critical situation people would like to know about, whereas the "operation timeout" (as it is called in the docs) can certainly be higher due to "normal use". To account for that, we could add another option (say `session_pool: [init_timeout: ...]` or so) instead of using the existing one. That way you could even write tests for both.
> 
> While probably a bit more work, I'd prefer option 2. What do you think?
> 
> cheers, malte

Went with the second option, not sure if this is a proper way to test the new option?

maltoe:
@dvic Thanks again! 🍰

andreasknoepfle:
Released in https://hexdocs.pm/chromic_pdf/1.2.0/readme.html

Issue #153, opened 2022-06-14
Possibly incompatible with Ghostscript 9.56.1


maltoe:
See https://elixirforum.com/t/chromicpdf-pdf-generator/29473/41

```
** (RuntimeError)   /usr/local/bin/gs exited with status 1!

GPL Ghostscript 9.56.1: Unrecoverable error, exit code 1


    (chromic_pdf 1.2.0) lib/chromic_pdf/utils.ex:53: ChromicPDF.Utils.system_cmd!/3
```

Haven't confirmed it yet.

andreasknoepfle:
Can confirm this happens on Ghostscript update:
```
 GPL Ghostscript 9.56.1: Unrecoverable error, exit code 1
 ```

**Comment by andreasknoepfle** [comment]
https://www.ghostscript.com/doc/current/History9.htm#Version9.56.0 

Maybe this: 
> In this (9.56.0) release, the new PDF interpreter is now ENABLED by default in Ghostscript, but the old PDF interpreter can be used as a fallback by specifying -dNEWPDF=false. We've provided this so users that encounter issues with the new interpreter can keep working while we iron out those issues, the option will not be available in the long term.

The other two incompatible changes are not really something that should affect us:
> (9.55.0) Changes to the device API. This will affect developers and maintainers of Ghostscript devices. Firstly, and most importantly, the way device-specific "procs" are specified has been rewritten to make it (we think!) clearer and less confusing. See [The Interface between Ghostscript and Device Drivers](https://www.ghostscript.com/doc/current/Drivers.htm) and [The Great Device Rework Of 2021](https://www.notion.so/artifexsoftware/The-Great-Device-Rework-Of-2021-94092fe1395d4a088b91462f0ca5038a) for more details.

> (9.55.0) The command line options -sGraphicsICCProfile=___, -dGraphicsIntent=#, -dGraphicsBlackPt=#, -dGraphicsKPreserve=# have been changed to -sVectorICCProfile=___, -dVectorIntent=#, -dVectorBlackPt=#, -dVectorKPreserve=#.

**Comment by andreasknoepfle** [comment]
It seems to have to do with quoted paths that we pass along to ghostscript. Will work on a fix for this.

**Comment by andreasknoepfle** [comment]
Fixed in #154

**Comment by andreasknoepfle** [comment]
There seems to be more problems with ghostscript 9.56.1. Now we see exit status 139. But it seems to be inconsistently happening.

```
[...]

PR #154, opened 2022-07-11
[#153] Possibly incompatible with Ghostscript 9.56.1

andreasknoepfle:
https://hexdocs.pm/elixir/1.12/System.html#cmd/3

> arguments do not need to be escaped or quoted for shell safety

The arguments seem to confuse `ghostscript`

### DIFF
--- a/lib/chromic_pdf/pdfa/ghostscript_impl.ex
+++ b/lib/chromic_pdf/pdfa/ghostscript_impl.ex
@@ -20,20 +20,20 @@ defmodule ChromicPDF.GhostscriptImpl do
   @impl ChromicPDF.Ghostscript
   def run_postscript(pdf_path, ps_path) do
     ghostscript_cmd!([
-      ~s(--permit-file-read="#{pdf_path}"),
-      ~s(--permit-file-read="#{ps_path}"),
+      ~s(--permit-file-read=#{pdf_path}),
+      ~s(--permit-file-read=#{ps_path}),
       "-dNODISPLAY",
       "-q",
-      ~s(-sFile="#{pdf_path}"),
-      ~s("#{ps_path}")
+      ~s(-sFile=#{pdf_path}),
+      ps_path
     ])
   end
 
   @impl ChromicPDF.Ghostscript
   def embed_fonts(pdf_path, output_path) do
     ghostscript_cmd!([
-      ~s(--permit-file-read="#{pdf_path}"),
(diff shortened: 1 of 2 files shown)


test/integration/pdfa_generation_test.exs
```diff
@@ -48,6 +48,20 @@ defmodule ChromicPDF.PDFAGenerationTest do
...
+      with_tmp_dir(fn tmp_dir ->
+        pdf_path = Path.join(tmp_dir, "output !@#$%.pdf")
+        pdfa_path = Path.join(tmp_dir, "output* !@#$%.pdf")
+
+        [content: "<p>Test</p>"]
+        |> ChromicPDF.Template.source_and_options()
+        |> ChromicPDF.print_to_pdf(output: pdf_path)
+
+        ChromicPDF.convert_to_pdfa(pdf_path, output: pdfa_path)
+        assert File.exists?(pdfa_path)
```
andreasknoepfle, line 61: I also checked, the file is generated perfectly.

lib/chromic_pdf/pdfa/ghostscript_impl.ex
```diff
@@ -7,7 +7,6 @@ defmodule ChromicPDF.GhostscriptImpl do
 
   @default_args [
     "-dQUIET",
-    "-sstdout=/dev/null",
```
maltoe, line 10: debugging artifact?

maltoe (approved):
:purple_heart: Nice!

lib/chromic_pdf/pdfa/ghostscript_impl.ex
andreasknoepfle: True, yes. But that is the reason why ghostscript is very silent when it comes to errors... 
But sure why we want to eat all output (especially since we direct stderr output to stdout). Do you know why we do this?

andreasknoepfle: We can improve that in another PR though. I'll create an issue if you like.

maltoe: Probably just wanted to avoid spamming the console with this.. IMO ideally we'd capture the logs and print them in case of an error only.

andreasknoepfle: Yes, but the above is only possible without the flag :D

PR #161, opened 2022-09-21
Add another known cause of crashing in Docker containers

WilliamVenner:
Hi there, recently I investigated an issue where, under stress, Chromium spawned by ChromicPDF would crash inside of the Docker container.

I eventually found that the `--disable-dev-shm-usage` flag fixes the issue. By default Docker containers have 64 MB of shared memory available in /dev/shm, so under stress, this shared memory is exhausted, and Chromium crashes: https://github.com/puppeteer/puppeteer/blob/v1.0.0/docs/troubleshooting.md#tips

`--disable-dev-shm-usage` fixes this by using /tmp instead, at the cost of potentially reduced performance since in most configurations /tmp will be disk I/O bound of course.

It may be worth adding this flag to the ChromicPDF API as an option, like with `no_sandbox`. But due to the performance concerns, whether this should be default or not is up for debate.

### DIFF
--- a/lib/chromic_pdf/pdf/protocol.ex
+++ b/lib/chromic_pdf/pdf/protocol.ex
@@ -132,6 +132,7 @@ defmodule ChromicPDF.Protocol do
       Known causes:
 
       1) External URLs in `<link>` tags in the header/footer templates cause Chrome to crash.
+      2) Shared memory exhaustion causes Chrome to crash. Docker containers only have 64 MB by default available in /dev/shm. Pass --disable-dev-shm-usage as a Chrome flag to use /tmp for this purpose instead, or increase the amount of shared memory available to the container with --shm-size.
       """)
     end
   end
--- a/lib/chromic_pdf/pdf/protocol.ex
+++ b/lib/chromic_pdf/pdf/protocol.ex
@@ -132,7 +132,7 @@ defmodule ChromicPDF.Protocol do
       Known causes:
 
       1) External URLs in `<link>` tags in the header/footer templates cause Chrome to crash.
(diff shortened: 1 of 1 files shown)


maltoe (approved):
Nice! Quite interesting find. How did you find out about this in the end?

maltoe:
> But due to the performance concerns, whether this should be default or not is up for debate.

Since you're the first to notice, I'd prefer to let it use shared memory by default. Ideally with a clear way for users to notice when they've run into the issue.

WilliamVenner:
> Thank you! Quite interesting. How did you find out about this in the end?

An issue was reported to our team about parallel requests timing out with ChromicPDF. It would seem that if you sent enough requests in parallel (20 or so) to be rendered as HTML->PDF, you could reproduce the crash. I tried an alternative librar and found that I could no longer reproduce the crash. Looking further into it, I discovered that the library was passing a bunch of arguments to Chrome by default, one of them being `--disable-dev-shm-usage`, so I isolated it down to that argument that fixed the crash, and added it to ChromicPDF chrome_args, and sure enough, it fixed the issue :)

> Another question: Is there a way for users to tell that they've run into this issue?

> Since you're the first to notice, I'd prefer to let it use shared memory by default. Ideally with a clear way for users to tell when they've run into the issue.

This very crash message will be logged, which makes this the perfect place for this tip 😆

Notably, I found virtually zero difference in the rendering time of a 1.4 MB HTML file spanning 117 pages with the flag on or off.

I would benchmark this with a load of parallel requests, but, unfortunately we know what would happen 😂

maltoe:
> This very crash message will be logged, which makes this the perfect place for this tip

:upside_down_face: I meant to differentiate it from other possible causes for crashes. But you're right, this is probably enough to inspire people to try without shm

Thank you for looking into this! :purple_heart:

PR #166, opened 2022-09-22
Dockerfile update

maltoe:
* Updated cowboy & plug_cowboy for compatibility (of integration tests) with OTP >= 2.4.0
    * New `cimg/elixir` image as base isn't good because a) new ubuntu
      doesn't allow to install chrome without snap, and b) circleci provides
      browsers via their "Orb" system (which is nice), but I'd like to be
      able to test on the image locally. There's also not much advantage to
      using the orb, given we use a non-cimg image anyway and hence circleci
      needs to pull for each job.
    * Switched to a debian-based image
    * Added README about how to build and test


.circleci/images/primary/README.md
```diff
@@ -0,0 +1,40 @@
...
+First acquire the credentials from @maltoe.
+
+```
+docker login
+docker push maltoe/chromic-pdf-primary:x.y.z
+```
+
+## Update the circleci config
+
+in `.circleci/config.yml`
```
andreasknoepfle, line 40: Maybe https://circleci.com/blog/using-circleci-workflows-to-replicate-docker-hub-automated-builds/ ?

maltoe: you mean to test that the image still builds? :shrug: don't think that's necessary

would be nice if one could somehow build and store it in some circleci repo, to stop circleci from downloading the image every time

andreasknoepfle: no I meant autobuilding if the image changes and using latest

andreasknoepfle (approved):

PR #167, opened 2022-09-26
Add a disable_scripts navigation option

MaeIsBad:
This is useful for security, preventing exploitation of v8 bugs, SSRF and many other potential issues.

An alternative to this implementation(calling `Emulation.setScriptExecutionDisabled`) that can be used with the current version of chromic_pdf is to set the chromium flag `--blink-settings=scriptEnabled=false`, but because of the lack of stability or documentation of chromium cmdline flags I think this is a way better solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 ### Added
-- Add disable_scripts option to `print_to_pdf/2`. 
+- Add disable_scripts option to `print_to_pdf/2`. See the security section of `lib/chromic_pdf.ex` for more details
 ## [1.2.2] - 2022-07-18
 
 ### Fixed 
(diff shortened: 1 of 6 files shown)


maltoe (approved):
:+1: Super nice addition! Thank you :purple_heart:  

Maybe some documentation would be nice, perhaps following [this section](https://github.com/bitcrowd/chromic_pdf/blob/ffc6d06bc3989445a04a53ce06f340fba89533c6/lib/chromic_pdf.ex#L48) in the security-related topics?

maltoe (approved):

MaeIsBad:
You're right, I think it's better to put this in SpawnSession too, I'll move that in a moment

maltoe:
Need to go now, will take a look at it later then. Thank again!

PR #168, opened 2022-09-26
Make disable_scripts into a global option

MaeIsBad:
See discussion in https://github.com/bitcrowd/chromic_pdf/pull/167


maltoe:
:purple_heart:

PR #169, opened 2022-09-26
update security considerations section

maltoe:
& refactor `:disable_scripts` a bit

### DIFF
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -34,21 +34,47 @@ defmodule ChromicPDF do
 
   ## Security Considerations
 
-  Before adding a browser to your application's (perhaps already long) list of dependencies, you
-  may want consider the security hints below.
+  By default, ChromicPDF will allow Chrome to make use of its own ["sandbox" process jail](https://chromium.googlesource.com/chromium/src/+/master/docs/design/sandbox.md).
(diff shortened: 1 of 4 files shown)


lib/chromic_pdf.ex
```diff
@@ -34,49 +34,59 @@ defmodule ChromicPDF do
...
-  By default, ChromicPDF will allow Chrome to make use of its own ["sandbox" process jail](https://chromium.googlesource.com/chromium/src/+/master/docs/design/sandbox.md).
-  The sandbox tries to limit system resource access of the renderer processes to the minimum
-  resources they require to perform their task.
-
-  However, in Docker containers running Linux images (e.g. images based on Alpine), and which
-  are configured to run their main job as a non-root user, this causes Chrome to crash on startup
-  as it requires root privileges to enter the sandbox.
+  In Docker containers running Linux images (e.g. images based on Alpine), and which
+  are configured to run their main job as a non-root user, the sandbox may cause Chrome to crash
+  on startup as it requires root privileges.
```
andreasknoepfle, line 89: Maybe leave an advice here what do do then.

andreasknoepfle (approved):

lib/chromic_pdf.ex
maltoe: It's below, doesn't show up because unchanged

PR #175, opened 2022-11-08
Fix rebar compilation caching

andreasknoepfle:
Maybe this helps with the error here #174 ?  plug_cowboy updated, which pulls in ranch.


maltoe (approved):
idk, maybe? :shrug: let's try

maltoe:
:D it doesn't even pass itself...

andreasknoepfle:
Seems like we need to store the deps folder for rebar as well (probably integration is enough) because it only symlinks things into deps when doing `mix compile`

See https://github.com/erlang/rebar3/issues/1824#issuecomment-464486224

PR #176, opened 2022-11-08
Replace plug_cowboy with bandit as integration test server.

maltoe:
This was easier than thought. Payback day.


andreasknoepfle:
🤷 I don't know. I'd leave it there, since it can happen with any elixir dep that depends on an erlang package.

andreasknoepfle:
No even sure that it is worth it switching to bandit, now that we know what is going on

maltoe:
> No even sure that it is worth it switching to bandit, now that we know what is going on

I know... but I figured why not try something new for a change.

andreasknoepfle (approved):

nickgnd:
Living on the edge 🚀

PR #181, opened 2022-12-02
Improve error handling of protocol execution

maltoe:
This patch improves the error handling of the session pool and the channel
    in various ways.

    - Introduce ExecutionError custom exception.
    - Make NimblePool.checkout/3 error more meaningful: The exit signal sent
      by NimblePool.checkout/3 is now caught and translated into a helpful
      ExecutionError.
    - Make the SessionPool.terminate_worker/3 callback gracefully terminate
      the target when the client failed with an error (channel timeout or
      inspectorCrashed now).
    - Minor changes to the timeouts in the session:
      - Make `@close_checkout` its own thing and set it to 1s.
      - Make `@checkout_timeout` explicit, but leave it at default 5s.
    - Turn the Channel timeout error into an `ExecutionError` as well.
    - Pull out the warning about the inspectorCrashed message from the
      protocol into the channel. This would previously print the warning as
      many times as protocols were being processed, which wasn't ideal.
      Moved to the channel where it's printed only once.
    - Refactor protocol a bit to not use a "result_fun", but simply return
      the result to the caller.
    - Introduce ChannelState module to modularize channel logic.
    - By tagging the protocols in the channel state with a ref, we can
      identify the protocol easily when the channel crashes. Use that to
      "cancel" the protocol on timeout.
    - Make the Channel timeout exception include an inspection of the actual
      protocol state (as now returned by "cancel_protocol") instead of the
      initial protocol parameter.
    - Unrelated drive-by: Allow evaluate option and wait_for option to be
      combined (used in tests to simulate delayed execution).
    - Unrelated drive-by: Move a few tests around and rename the
      "SessionRestartTest" to "SessionPoolTest"

## TODO

- [x] add a test case that demonstrates that cancellation works


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## Unreleased
+
+### Added
+
+- Add error details on failure to checkout worker from pool.
+
+### Changed
+
+- `wait_for` and `evaluate` options can now be combined, as the `wait_for` script is appended to the user-given `evaluate` expression.
```
andreasknoepfle, line 9: Can you extract this into a separate PR ? It is already hard to review without having 2 in one :D

maltoe: Not really, I used this specifically to make the test work in the way you suggested (using a script that delays the `ready-to-print` thing)

lib/chromic_pdf/pdf/browser/session_pool.ex
```diff
@@ -148,18 +187,31 @@ defmodule ChromicPDF.Browser.SessionPool do
...
+  def terminate_worker(:shutdown, _worker_state, pool_state) do
+    {:ok, pool_state}
+  end
+
+  # Unexpected error reasons: :DOWN | :timeout | :throw | :exit
+  def terminate_worker(reason, _worker_state, pool_state) do
+    IO.warn("ChromicPDF worker terminated due to unexpected reason: #{reason}")
+
     {:ok, pool_state}
   end
```
andreasknoepfle, line 216: Lets remove this

lib/chromic_pdf/pdf/browser/channel_state.ex
```diff
@@ -0,0 +1,66 @@
...
+  alias ChromicPDF.Protocol
+
+  @enforce_keys [:dispatch, :protocols]
+  defstruct [:dispatch, :protocols]
+
+  @type t :: %__MODULE__{dispatch: Protocol.dispatch(), protocols: [Protocol.t()]}
+
+  @spec new(Protocol.dispatch()) :: t()
+  def new(dispatch) do
+    %__MODULE__{dispatch: dispatch, protocols: []}
```
andreasknoepfle, line 16: tbh, I don't know what `dispatch` is even for. I only find one reference in `Protocol.advance/2` here, which seems to me to be ignored everywhere, right? 

If so wouldn't it be nicer if this module would be called `ProtocolExecution` and the state would be an array of them (with `defstruct [:protocol, :from, :ref]`) ? Some bits would move back to the channel, but I think this could work nicely still.

lib/chromic_pdf/pdf/protocol.ex
```diff
@@ -40,110 +40,65 @@ defmodule ChromicPDF.Protocol do
...
-
-    protocol
-    |> test(msg)
-    |> advance(dispatch)
+  def advance(%__MODULE__{steps: [{:output, output_fun}], state: state}, _dispatch) do
+    {:halt, {:ok, output_fun.(state)}}
   end
 
-  defp test(%__MODULE__{steps: steps, state: state} = protocol, msg) do
+  @spec msg_in(__MODULE__.t(), JsonRPC.message()) :: __MODULE__.t()
```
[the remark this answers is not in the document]
maltoe: :+1: It would be the first mention of "RPC" in the codebase (though arguably it is just that). How about `devtools_message`? or `chrome_msg`?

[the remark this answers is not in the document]
andreasknoepfle: `chrome_message`& `handle_chrome_message` seem good choices. Abbreviations read ugly imho :D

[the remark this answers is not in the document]
maltoe: oh... you're right, RPC is actually in there a few times :see_no_evil:

lib/chromic_pdf/pdf/browser/channel_state.ex
maltoe: `dispatch` is used in `Protocol.advance/2` when it processes `:call` steps. It allows the protocols to send messages (as a side-effect). I can refactor that to instead return a list of messages instead?

I'm not sure about `ProtocolExecution` still, or maybe I don't fully understand what you're saying: If it's just about making the triple in `ChannelState` a map/struct, I'm all for it, but I would prefer not to introduce either `from` or `ref` into `Protocol.t`

maltoe: how about `ChannelTask` or `ChannelOperation`?

andreasknoepfle: both work I guess... Slightly in favour of `ChannelTask`

>  I can refactor that to instead return a list of messages instead? 

Seems like a nice choice.

lib/chromic_pdf/pdf/browser/channel.ex
```diff
@@ -50,34 +47,117 @@ defmodule ChromicPDF.Browser.Channel do
...
+
+  # Pops a value for which predicate is truthy from waitlist, returns predicate result.
+  defp pop_from_waitlist(%{waitlist: waitlist} = state, fun) do
+    {matched, waitlist} = do_pop_from_waitlist(waitlist, fun, [])
+    {matched, %{state | waitlist: waitlist}}
   end
 
-  defp update_protocols(state, protocols) do
-    %{state | protocols: Enum.reject(protocols, &Protocol.finished?(&1))}
+  defp do_pop_from_waitlist([], _fun, acc), do: {nil, acc}
```
andreasknoepfle, line 154:
```suggestion
  defp pop_if([], _fun, acc), do: {nil, acc}
```
it actually has nothing to do with a waitlist (would ne a great addition to the `List` module :D

lib/chromic_pdf/pdf/browser/channel.ex
```diff
@@ -50,34 +47,117 @@ defmodule ChromicPDF.Browser.Channel do
...
-        {:run_protocol, protocol},
-        from,
-        %__MODULE__{dispatch: dispatch, protocols: protocols} = state
-      ) do
-    protocol = Protocol.init(protocol, &GenServer.reply(from, &1), dispatch)
-
-    {:noreply, update_protocols(state, [protocol | protocols])}
+  @impl GenServer
+  def handle_call({:run_protocol, protocol, ref}, from, state) do
+    task = %{protocol: protocol, from: from, ref: ref}
```
andreasknoepfle, line 56: Kinda liked the `ChannelTask` 🤷 but also fine to keep it as a private thingy in the channel

lib/chromic_pdf/pdf/browser/channel.ex
andreasknoepfle: Ahh realized that it is in fact not `pop_if` but returns the predicate.... 🙈 forget it

andreasknoepfle (approved):
Great work!

PR #193, opened 2022-12-09
Add warmup/1

maltoe:


lib/chromic_pdf/pdf/chrome_runner.ex
```diff
@@ -1,25 +1,66 @@
...
     send(port, {self(), {:command, msg <> "\0"}})
 
     :ok
   end
 
-  # --------------------------- Chrome Command Line ----------------------------------
-  #
+  @spec warmup(keyword()) :: :ok
+  def warmup(opts) do
+    "<html><head></head><body></body></html>\n" =
```
andreasknoepfle, line 28:
```suggestion
    "<html>" <> _ =
```
does it matter what is printed? Seems fragile. When chrome decides to remove the additional `\n` from `about:blank` then this would break, right?

andreasknoepfle (approved):

lib/chromic_pdf.ex
```diff
@@ -299,4 +299,22 @@ defmodule ChromicPDF do
   """
 
   use ChromicPDF.Supervisor
+
+  @doc """
+  Runs a one-off Chrome task to warm up Chrome's caches.
+
+  TODO
```
andreasknoepfle, line 306: TODO

andreasknoepfle (approved):
Lgtm, little comments.

test/integration/warmup_test.exs
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
+
+defmodule ChromicPDF.WarumupTest do
```
markquezada, line 3:
```suggestion
defmodule ChromicPDF.WarmupTest do
```

lib/chromic_pdf/pdf/chrome_runner.ex
maltoe: Just noticed it's actually broken when using `discard_stderr: false`

```
iex(1)> ChromicPDF.warm_up(discard_stderr: false)
** (MatchError) no match of right hand side value: "[1214/100758.006638:WARNING:bluez_dbus_manager.cc(247)] Floss manager not present, cannot set Floss enable/disable.\n<html><head></head><body></body></html>\n"
    (chromic_pdf 1.4.0) lib/chromic_pdf/pdf/chrome_runner.ex:28: ChromicPDF.ChromeRunner.warm_up/1
```

maltoe: This is actually pretty surprising IMO

```elixir
iex(1)> :os.cmd(String.to_charlist("echo foo >&2"))
'foo\n'
```

Docs on [`os.cmd`](https://www.erlang.org/doc/man/os.html#cmd-1) say 

> Executes Command in a command shell of the target OS, captures the standard output of the command, and returns this result as a string.

... which would also be my intuition for its behaviour. But nope, it also captures `stderr` :shrug:

maltoe: See also http://erlang.org/pipermail/erlang-questions/2016-June/089640.html

maltoe:
@andreasknoepfle could you give this another :eyes: :pray:

lib/chromic_pdf/pdf/chrome_runner.ex
maltoe: changed this around a bit, removing the stdout output (TBH I don't think `about:blank` is going to change much in terms of markup) and returning the capture stderr

test/integration/warm_up_test.exs
```diff
@@ -0,0 +1,19 @@
...
+    test "fires a one-off chrome task to warmup Chrome's caches" do
+      assert ChromicPDF.warm_up() == {:ok, ""}
+    end
+
+    test "returns stderr when given discard_stderr: false" do
+      # Chrome crashes when given --dump-dom and more than one positional (non --) argument.
+      {:ok, stderr} =
+        ChromicPDF.warm_up(discard_stderr: false, chrome_args: "another_positional_arg")
+
+      assert stderr =~ "Open multiple tabs is only supported when remote debugging is enabled"
```
andreasknoepfle, line 16: Is this consistent on mac & linux chrome and chromium?

andreasknoepfle: tested on mac at least and I know you cover linux, so 👍

andreasknoepfle (approved):

test/integration/warm_up_test.exs
maltoe: thank you :purple_heart:

Issue #177, opened 2022-11-16
Rendering a combined PDF of multiple "files"


markquezada:
This is more of a question I suppose. I currently generate a PDF for a specific user based on their account information. 

I have the need to generate a single PDF of a group of these individual PDFs for archival purposes. Currently I'm rendering them individually to disk and then concatenating them via a `System.cmd` call to `qpdf`.

I thought about rendering one long PDF but I'm currently using both the header (for user profile info) and footer (for page numbers) to generate the pdf so I don't believe that'll work since headers and footers span the entire document.

Is there any way for me to leverage chromicPDF to generate this combined file directly?

maltoe:
Hi @markquezada 

No, Chrome won't be of any help. I suppose you can join PDFs with Ghostscript (which you may have installed?), but TBH it's probably easier to use `qpdf` or `pdfjoin` and do it in a postprocessing step as you do. Sorry.

Best,
malte

markquezada:
Figured, thanks!

maltoe:
@markquezada FYI #194

PR #194, opened 2022-12-14
Add merge feature

xaviRodri:
In order to be able to merge multiple PDF files, we have added this Ghostscript command that can be called using `ChromicPDF.merge/2`.

This new function accepts a list of paths and an output option that can be added or not depending on the desired output (same as `print_to_pdf/2` output option).


xaviRodri:
Thanks for the fast answer @maltoe !! 💪 

First, to answer to your questions:

> What happens to the metadata set on the individual PDFs?

It depends on 2 factors:

* if only one of them has metadata, the merged one will take it from that one (order doesn't affect here)
* if more than one has metadata, the last one being merged will give the meta to the merged file (order is important here)

> I'm assuming the PDF/A features are also "lost" when merging two PDF/A files. Did you try that?

I'm not super familiar with all the PDF/A features.. but I tried to see an example file with `xpdf` and looks like the merged file still warns me about syntax and so.

Also, I like a lot the idea to get a list of source and options and create a `print_to_pdf_and_merge/2` function!
Let me try it too because that would fit my use case 100%

Thanks again! 💯

maltoe:
@xaviRodri Thought about it some more and figured we could as well just extend the existing `print_to_pdf/2` function with a new clause. I pushed a small commit changing around the parameter types slightly, hopefully you can make sense of it.

New signature would be:

```
@spec print_to_pdf([source() | source_and_options()], [export_option()]) :: export_return()
```

Behaviour:

* create a tmpdir
* print a pdf for each source in the list
* merge them all and return the file as indicated by the `output` option (so follow the existing API design)

Things to look out for:

* telemetry should be called only once, so you would need to rearrange the `chrome_export` utility a bit (sorry for the mess, btw, I know the API implementation is a bit confusing)
* documentation needs to be clear on the fact that this requires ghostscript (as `print_to_pdf` otherwise doesn't)

-> With this new function I think we should not add a top-level `ChromicPDF.merge/2` API anymore.

---

Next, since we're using ghostscript for the merge anyway, we would be able to use the existing ghostscript functionality, i.e. adding metadata (see `t:info_option`) as well as PDF/A conversion. But if you think this is too much for a single PR, feel free to use the `Ghostscript.merge/2` function you've already implemented for now, and we'll see further afterwards.

xaviRodri:
Thanks for taking a look!!

I'm right now trying to get it work 👍 

what do you think about this spec?

```Elixir
@spec print_to_pdf(source() | source_and_options() | [source() | source_and_options()], [export_option()]) :: export_return()
```

So the current people using it don't need to change it into a list?
I can try to patternmatch or add a guard so we can distinguish the different options 👌

xaviRodri:
Some updates @maltoe !

I got it working! 🎊 
Currently I have it as a separate function but takes the same arguments as `print_to_pdf/2`.

I'm running into a "problem" that I've solved but maybe some help could improve the solution..

So, right now, if we use `source_and_options/1` to create a source and then we call `print_to_pdf/2`, what we are doing is overriding the source's options (let's call them individual opts) with the function options (general opts maybe?).

But I was thinking on the possibility of merging both instead of overriding them, so we would be able to have some general attributes (e.g. margins for the whole merged pdf) but at the same time specify some for a particular "section" of the final file (e.g. I want the top margin bigger in this section, I want a different header for that section, etc.).
This won't work if we just override the `print_to_pdf` option... So I created a `merge_print_options/2` function that, for each source, merges some of the options we have there.
I had conflicts with some default values that doesn't appear in the result of `source_and_options/1`, so I had to do some changes to the `Template` module.

Let me push these draft changes so you can check the idea better (don't expect a super clean code, specs and so as it needs still some work!)
If you think that these may be too much for what we need let me know! I'm 100% open to change it as many times as needed.

lib/chromic_pdf/api/pdf_options.ex
```diff
@@ -118,4 +118,23 @@ defmodule ChromicPDF.PDFOptions do
...
+
+      nil ->
+        data =
+          pdf_path
+          |> File.read!()
+          |> Base.encode64()
+
+        {:ok, data}
+    end
+  end
```
maltoe, line 139: I think this is a copy of `PDFAOptions.feed_ghostscript_file_into_output/2`

Let's move this code to a different module, perhaps `ChromicPDF.Utils.feed_file_into_output/2` or so. Or into a new module in the `api/` directory, perhaps `OutputOption`?

lib/chromic_pdf/template.ex
```diff
@@ -171,7 +176,7 @@ defmodule ChromicPDF.Template do
     content = Keyword.get(opts, :content, @default_content)
     header = Keyword.get(opts, :header, "")
     footer = Keyword.get(opts, :footer, "")
-    styles = do_styles(opts)
+    styles = if(opts[:style] == :custom, do: "", else: do_styles(opts))
```
maltoe, line 179: No, sorry, this is not what we want, IMO. The whole idea of `Template` is to abstract the complexity of dealing with page dimensions in CSS and in the `printToPDF` options, so that as a user I can simply say "I want A4 paper and have a header of 4cm". This is solved by prepending this custom stylesheet to the input, so if you remove the stylesheet, using `Template` at all is kind of pointless (except maybe the `:a4` -> inch conversion, but maybe we can expose that as a function instead, if its useful by itself).

If a user does not want to use this behaviour, they're free to set their own `printToPDF` options and not use `Template`. They can also construct `source_and_options` structs if they now want to set `printToPDF` options for each of their multiple inputs. We can explain this in the docs when we describe multiple inputs handling.

Last but not least, if in your specific use-case you need to set additional page margins (especially left/right), you can easily add them in your own stylesheet. That's at least how we internally use `ChromicPDF.Template`: It should set up the overall page size (ISO A4) and tell Chrome how big the header is, but otherwise don't apply any margins, as we set them in the stylesheets rendered with our templates later.

maltoe:
@xaviRodri Nice work already!

lib/chromic_pdf/api.ex
```diff
@@ -42,6 +55,52 @@ defmodule ChromicPDF.API do
     print_to_pdf: PrintToPDF
   }
 
+  @margin_opts [:marginTop, :marginBottom, :marginLeft, :marginRight]
+
+  defp merge_print_options(sources, opts) when is_list(sources) do
+    Enum.map(sources, &merge_print_options(&1, opts))
+  end
+
+  defp merge_print_options(%{opts: [print_to_pdf: single_opts] = source_opts} = source, opts) do
```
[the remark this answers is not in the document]
xaviRodri: I totally agree on simplicity 💯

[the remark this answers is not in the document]
xaviRodri: > Adding a default option (say a marginLeft) to each of your inputs is trivial to solve in user's application logic

how could we do that if we want the generated PDFs to be merged?
I didn't find how to, and that's why I made this

lib/chromic_pdf/template.ex
xaviRodri: No problem at all! 👍 
It was only a workaround to show you the status of it (I also didn't like it much)

In my case I was more worried of Top/Bottom margins as I have to play with them for different headers/footers...
but I will set individual `header_height` & `footer_height` for all the templates 👌

xaviRodri:
I will play on it today too so I can address all the above 🔝 

But I think we are now on the good direction 💯 

Thanks a lot for the feedback till now 🙏

lib/chromic_pdf/template.ex
maltoe: Exactly :+1: Idea of `Template` is to abstract the `printToPDF` opts, as Chrome's behaviour (in particular in combination with `@page` stylesheets) is somewhat peculiar.  That's also why I named these options `header_height` and `footer_height`, to separate them from the underlying API.

xaviRodri:
More updates! 👀 

I have a couple of questions before writing documentation & tests:

1. What to do with the `with_telemetry` clauses - I remembered you said we should only call it once.. but once for all the process or once for all the `chrome_export`s we do? (e.g. in `print_to_pdfa/3` it's call it twice) I could move it one level up so `with_telemetry` calls the `chrome_export` (that's maybe what you wanted to say with "rearrange the chrome_export utility a bit").
2. Should I maintain the API's `merge` feature? Or should we only have it as a tool for the "print & merge"?

For the rest, if you have a sec to check it, I'm happy to change anything you find weird 👍

lib/chromic_pdf/api.ex
```diff
@@ -10,16 +10,36 @@ defmodule ChromicPDF.API do
...
+          | ChromicPDF.source_and_options()
+          | [ChromicPDF.source() | ChromicPDF.source_and_options()],
           [ChromicPDF.pdf_option() | ChromicPDF.export_option()]
         ) :: ChromicPDF.export_return()
+  def print_to_pdf(services, sources, opts) when is_list(sources) and is_list(opts) do
+    with_tmp_dir(fn tmp_dir ->
+      sources =
+        Enum.map(sources, fn
+          %{opts: source_opts} = source -> %{source | opts: Keyword.merge(source_opts, opts)}
+          source -> source
```
maltoe, line 30: The way the options are passed through to `print_tmp` are a bit confusing, IMO. Why not just

```suggestion
          source -> %{source: source, opts: opts}
```

and remove the last argument of `print_tmp/4`?

lib/chromic_pdf/api.ex
```diff
@@ -83,12 +115,23 @@ defmodule ChromicPDF.API do
...
+  @spec merge(ChromicPDF.Supervisor.services(), list(ChromicPDF.path()), keyword()) ::
+          ChromicPDF.export_return()
+  def merge(services, pdf_path_list, opts) do
+    with_tmp_dir(fn tmp_dir ->
+      tmp_path = Path.join(tmp_dir, random_file_name(".pdf"))
+
+      :ok = GhostscriptPool.merge(services.ghostscript_pool, pdf_path_list, opts, tmp_path)
+      OutputOptions.feed_file_into_output(tmp_path, opts)
+    end)
+  end
```
maltoe, line 127: let's remove this for now

lib/chromic_pdf/api/output_option.ex
```diff
@@ -0,0 +1,52 @@
...
+            fun.(path)
+          end)
+
+        {:ok, result_from_callback}
+
+      nil ->
+        {:ok, data}
+    end
+  end
+end
```
maltoe, line 52: :+1:

lib/chromic_pdf/pdfa/ghostscript_worker.ex
```diff
@@ -35,6 +35,15 @@ defmodule ChromicPDF.GhostscriptWorker do
     :ok
   end
 
+  @spec merge(list(binary()), keyword(), binary()) :: :ok
+  def merge(pdf_path_list, _params, output_path) do
+    pdf_path_list = Enum.map(pdf_path_list, &Path.expand/1)
```
maltoe, line 40: why is this step necessary?

maltoe:
@xaviRodri 

> e.g. in print_to_pdfa/3 it's call it twice

Hmm, wasn't aware of this anymore. OK then, maybe we align with the existing behaviour and continue to send one `print_to_pdf` event for each print job and one new `merge` event for the merger? Not sure :shrug: IMO having one combined events would make it more useful for instrumentation, but we shouldn't change existing behaviour.

> Should I maintain the API's merge feature? 

Let's get rid of it for now. If someone wants to only merge PDFs, they can use `pdfjoin` or the like. Keep the main API small.

lib/chromic_pdf/api.ex
xaviRodri: good point here! 👍

lib/chromic_pdf/api/output_option.ex
xaviRodri: Glad you find it good! 💯

lib/chromic_pdf/pdfa/ghostscript_worker.ex
xaviRodri: True! Not necessary at all.
It's something I added at the beginning because I was following the example from `convert/3` 😅 
Deleting it 💥

maltoe:
@xaviRodri Looking good, thanks!

When you document this, please make sure to mention that Ghostscript is required for the merge step and that printing of individual files before merge is done sequentially. For testing, I'd suggest to add a test case to `pdf_generation_test.exs` and perhaps one to `telemetry_test.exs`

andreasknoepfle:
> printing of individual files before merge is done sequentially

I wonder if it would be worth adding a ticket after this to allowing parallel printing here (at least up to the pool size?)

maltoe:
> allowing parallel printing

I think that is more complex than we think, but yeah, we could explore it for sure.

xaviRodri:
Hey again!

Added some tests and documentation as we agreed!
I'm not 100% sure if the tests cover all the expected for this merge part, but I didn't want to add tests for already tested features (dynamic templates, etc.).

Also about the documentation: I added it inside "Input options" as this is related to the inputs. But if you think it should belong to another part let me know! 👍 

Marking it as "Ready for review"! 🚀

test/integration/pdf_generation_test.exs
```diff
@@ -206,6 +206,55 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+        end
+      )
+    end
+  end
+
+  describe "multiple PDF printing & merge" do
+    setup do
+      start_supervised!(ChromicPDF)
+      :ok
+    end
```
maltoe, line 257: You probably wanted this `describe` section to contain the tests above

maltoe (approved):
Thanks, this is great!

Since we already had a few iterations - and since I realized that I want to refactor a few things (types, docs, Ghostscript module) -, I'd merge this as-is and "own" your code in a follow-up commit. Hope you don't mind.

Thanks a lot for the contribution! :purple_heart: :blue_heart: :yellow_heart: Super happy about how this turned out. Very nice addition, especially given we've recently been asked about this very feature in #177

test/integration/pdf_generation_test.exs
xaviRodri: First I started in here but moved them i to the other describe!
Sorry I forgot to delete this 😅

lib/chromic_pdf/supervisor.ex
```diff
@@ -373,6 +373,42 @@ defmodule ChromicPDF.Supervisor do
           content = SomeView.render("body.html") |> Phoenix.HTML.safe_to_string()
           ChromicPDF.print_to_pdf({:html, content})
 
+      ### Multiple source input & merge
+
+      **This feature requires Ghostscript to merge different files.**
+
+      If you have multiple inputs, you can pass them as a list of sources. In this case, the different results will be merged into one single output.
+      Remember that the merge will be done sequentially, so the order of the given list is important.
```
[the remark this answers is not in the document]
xaviRodri: Oh I missunderstood it then!
Time of printing is also something to be aware of, true!
Let’s see if what @andreasknoepfle said about concurrent printing can become real 💪

PR #198, opened 2022-12-22
Enable multiple sources for print_to_pdfa/2

maltoe:


test/integration/pdfa_generation_test.exs
```diff
@@ -50,6 +50,10 @@ defmodule ChromicPDF.PDFAGenerationTest do
       end
     end
 
+    test "it joins multiple sources into a single PDF" do
+      assert {:ok, _} = ChromicPDF.print_to_pdfa([{:html, "page 1"}, {:html, "page 2"}])
```
andreasknoepfle, line 54: should we add a test that the pdfa parts actually work?

PR #204, opened 2022-12-22
Refactor Ghostscript command line args building

maltoe:
- Remove unnecessary embed fonts step, simply embed the fonts in the conversion. Tried this with a webfont from a `:url` and it still works. Not sure what made me initially use a separate step for this. Cargo cult programming.
- Instead moved the relevant arguments to a new list of default pdfwrite arguments. Some of them are the default, anyway (like `dEmbedAllFonts`).
- Refactored lowest-level `GhostscriptImpl` wrapper
  - Drop interface and & `Impl` suffix, align with `ChromeRunner`
  - Takes care of command line arguments only
  - Replace `convert_to_pdfa` function with `pdfwrite` function that takes a list of input files and optionally applies the PDF/A
  - Remove `join` function and use `pdfwrite` instead
  - Dropped conditional handling of safer mode after GS 1.50 for unconditional `dSAFER` flag
- Made sure we apply escaping to paths everywhere


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Changed
```
maltoe, line 3: could also be `Fixed`, not sure :shrug:

maltoe:
@ammancilla once this is merged, we could release another 1.5.1 or 1.6.0 or so and then start breaking API for a 2.0

lib/chromic_pdf/utils.ex
```diff
@@ -62,6 +62,10 @@ defmodule ChromicPDF.Utils do
     end
   end
 
+  @spec shellescape(binary()) :: binary()
+  def shellescape(value),
```
andreasknoepfle, line 66: pls kick this

andreasknoepfle (approved):

andreasknoepfle:
Looks great

PR #206, opened 2022-12-23
Fix typos

kianmeng:
Found via `codespell -S deps -L warmup,ot`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,13 +42,13 @@
 
 ### Fixed
 
-- Use old pdf interpreter with ghostcript > 9.56 to avoid segfaults (see #158)
+- Use old pdf interpreter with ghostscript > 9.56 to avoid segfaults (see #158)
 
 ## [1.2.1] - 2022-07-12
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


andreasknoepfle:
@maltoe maybe we should switch to action at some point. I have no clue, why CircleCI keeps forgetting out configs...

andreasknoepfle (approved):

maltoe (approved):

maltoe:
> @maltoe maybe we should switch to action at some point. I have no clue, why CircleCI keeps forgetting out configs...

Agree. Not even too difficult anymore, IMO

- [ ] Find out why this one test case always fails ([example run](https://github.com/bitcrowd/chromic_pdf/actions/runs/3704537014/jobs/6277273477))
- [ ] Find a way to install verapdf & ZUV without doing it manually every time (or ditch ZUV). Is it possible to use docker images in actions?

maltoe:
Also, thanks @kianmeng 

@andreasknoepfle we should also make `codespell` part of our pipelines

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

Issue #207, opened 2022-12-27
No accessibility tags in Ghostscript-generated files


MaeIsBad:
Hey! Sorry for not taking a look at this when the PR was originally in review but there are a few things you should know related to merging chromic_pdf generated documents with ghostscript:

- it's really slow. To quote an ex-coworker
> In my testing, I found that, with a large PDF file (~200 pages), GhostScript took 12 seconds to handle the request, and pdfunite took... 4.47 seconds! This should therefore speed up merging PDFs significantly. Turns out, ghostscript is a significant and slow bottleneck.

- it removes the accessibility annotations added via `--export-tagged-pdf`. This seems like something that you don't want happening implicitly
https://github.com/bitcrowd/chromic_pdf/blob/f8d7a391f649b861579b6432041d04ea512e0546/lib/chromic_pdf/pdf/chrome_runner.ex#L109

maltoe:
Hey @MaeIsBad 

thank you for your comments. I was unaware of the issue that Ghostscript is slow to join multiple PDFs, though am not surprised. But the factor 2  (in comparison to a specialized tool like pdfunite) you mention and also the 12s / 200 pages measurement don't seem excessive to me, so I would guess that for most users this is not a big deal in reality.

I'm aware that Ghostscript drops all structural information from a source PDF when writing to its `pdfwrite` output device, unfortunately. This also affects the PDF/A conversion. In our own apps, accessibility of the generated files (mostly invoice documents or the like) has never been a priority - not saying we wouldn't prefer files to keep their structural information, only that PDF/A features like fixed presentation are more critical to us. IOW, since (luckily) we need `PDF/A-3b` compliance only, not `-3a`, we didn't really bother looking into this topic.

According to [internet wisdom](https://stackoverflow.com/q/69299263), Ghostscript is not capable of transferring these annotations from input documents to the output, as in, `pdfwrite` can write pdfmarks perfectly fine, only they're not forwarded from the PDF interpreter. I personally wonder if this is going to change with [their new PDF interpreter](https://ghostscript.com/blog/pdfi.html), but haven't had the time to look into this yet.

As a follow-up, I'd suggest we:

- Improve the documentation:
  - the "Multiple Sources" section of `print_to_pdf/2` should clearly mention that Ghostscript removes tags, and that the user should use other tools if accessibility of the files is a requirement
    -  btw, `pdfunite` also doesn't retain the tags, so I don't know any tool that does, atm
[...]

PR #208, opened 2022-12-29
[#207] Document state of accessibility with Chrome/Ghostscript

maltoe:
Addresses #207

### DIFF
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -293,9 +293,10 @@ defmodule ChromicPDF do
   default](https://blog.chromium.org/2020/07/using-chrome-to-generate-more.html).  These files
   contain structural information about the document, i.e. type information about the nodes
   (headings, paragraph, etc.), as well as metadata like node attributes (e.g., image alt texts).
-  This information allows assistive tools like screen readers to do their job. `pdfinfo`
-  identifiers these files as `Tagged: yes`, and you may review some of the contained information
-  with the `pdfinfo -struct-text <file>` command.
+  This information allows assistive tools like screen readers to do their job, at the cost of
+  (at times significantly) increasing the file size. To check whether a PDF file is tagged, you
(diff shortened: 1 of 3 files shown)


lib/chromic_pdf.ex
```diff
@@ -279,28 +287,29 @@ defmodule ChromicPDF do
...
-    Initiative. If you need to be able to use a different color scheme, please open an issue.
+  ## On Accessibility / PDF/UA
+
+  Since its version 85, Chrome generates "Tagged PDF" files [by
+  default](https://blog.chromium.org/2020/07/using-chrome-to-generate-more.html).  These files
+  contain structural information about the document, i.e. type information about the nodes
+  (headings, paragraph, etc.), as well as metadata like node attributes (e.g., image alt texts).
+  This information allows assistive tools like screen readers to do their job. `pdfinfo`
+  identifiers these files as `Tagged: yes`, and you may review some of the contained information
+  with the `pdfinfo -struct-text <file>` command.
```
MaeIsBad, line 298: This is an extreme example but regularly we have to generate a massive(200+ page) pdf consisting of mostly tables. When tagged the resulting PDF weighs 40MiB, while the untagged version is just 3.1MiB. 
I'm not sure if this is something that's worth mentioning in the docs since it's an extreme case that already causes chromium to take minutes generating the document, and we really shouldn't be using PDFs.

MaeIsBad (approved):
Very detailed and well written docs :pray: great job

lib/chromic_pdf.ex
maltoe: mentioned in c48ba6db3bd91e148657b68cacd804aee072ceb8

Is it actually still possible to tell Chrome *not* to include these annotations? i.e. `--no-export-tagged-pdfs`?

MaeIsBad: `--disable-pdf-tagging`

PR #211, opened 2023-01-11
Github Packages images workflow

agatheblues:
The PR introduces a workflow that automatically builds images and push them to Github Packages.

References:
- https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages
- https://docs.github.com/en/packages/quickstart 
- https://thedatabaseme.de/2022/07/02/building-and-pushing-use-github-actions-to-build-and-push-docker-images/

The packages are visible on the main page of the repo under the `releases` section on the right.

image


.circleci/images/primary/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN wget http://downloads.verapdf.org/rel/verapdf-installer.zip \
   && unzip verapdf-installer.zip \
   && mv verapdf-greenfield* verapdf-greenfield \
   && chmod +x verapdf-greenfield/verapdf-install
-COPY auto-install.xml /opt/verapdf/verapdf-greenfield
+COPY .circleci/images/primary/auto-install.xml /opt/verapdf/verapdf-greenfield
```
maltoe, line 26: Guess we should move the images to `.github/images` or so, right?

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,56 @@
...
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          file: .circleci/images/primary/Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
maltoe, line 56: not sure I understand this `docker/metadata-action` thing. Can we just call the images something like `chromic-pdf-test-image-debian-bullseye:latest` or so?

maltoe (commented):
:clap: :clap: :clap:

.github/workflows/publish-image.yml
maltoe: maybe incl. the ghostscript / chrome versions

.github/docker/Dockerfile_alpine-3-16-2
```diff
@@ -0,0 +1,40 @@
...
+RUN wget https://github.com/ZUGFeRD/ZUV/releases/download/v0.8.3/ZUV-0.8.3.jar
+ENV ZUV_JAR /opt/zuv/ZUV-0.8.3.jar
+
+RUN apt-get remove -y \
+    wget \
+    unzip \
+  && apt-get autoremove -y \
+	&& rm -rf /var/lib/apt/lists/*
+
+RUN mix local.hex --force && mix local.rebar
```
maltoe, line 40: i wonder if running these as `root` will work (when we run `mix deps.get` & `mix compile` as non-root)

maltoe (approved):
Amazing!

PR #212, opened 2023-01-11
Test on Github Actions with custom image

maltoe:
This patch configures a new Github Actions workflow called "test" which
    makes use of our new test images stored in Github Packages. A new
    workflow called "lint" does the linting, the CircleCI pipeline is dropped.

    As a consequence from testing on Alpine 3.16, we do now support
    Ghostscript 9.56 properly.


agatheblues:
We should delete the `.circleci` folder then?

maltoe:
> We should delete the `.circleci` folder then?

eventually

agatheblues:
🧹  ✨

agatheblues (approved):

Issue #219, opened 2023-01-19
Handle DOWN in terminate_worker/3


markquezada:
Hi, 

We just started running into this issue in production:

> 2023-01-19T20:31:56.032465131Z 20:31:55.994 [error] Error during ChromicPDF.Browser.SessionPool.terminate_worker/3 callback:
> 2023-01-19T20:31:56.032514431Z ** (FunctionClauseError) no function clause matching in ChromicPDF.Browser.SessionPool.terminate_worker/3
> 2023-01-19T20:31:56.032520231Z     (chromic_pdf 1.6.0) lib/chromic_pdf/pdf/browser/session_pool.ex:191: ChromicPDF.Browser.SessionPool.terminate_worker(:DOWN, %{session: %{session_id: "B4192E860C6DB1C83D3A4404A257EFD0", target_id: "B3EDCFD2C0E3D3349617B0A1F59AEC0F"}, uses: 14}, %{browser: #PID<0.2847.0>, init_timeout: 5000, max_session_uses: 1000, spawn_protocol: %ChromicPDF.Protocol{steps: [call: &ChromicPDF.SpawnSession.create_browser_context/2, await: &ChromicPDF.SpawnSession.browser_context_created/2, call: &ChromicPDF.SpawnSession.create_target/2, await: &ChromicPDF.SpawnSession.target_created/2, call: &ChromicPDF.SpawnSession.attach/2, await: &ChromicPDF.SpawnSession.attached/2, call: &ChromicPDF.SpawnSession.set_user_agent/2, call: &ChromicPDF.SpawnSession.offline_mode/2, call: &ChromicPDF.SpawnSession.enable_page/2, await: &ChromicPDF.SpawnSession.page_enabled/2, call: &ChromicPDF.ResetTarget.reset_history/2, await: &ChromicPDF.ResetTarget.history_reset/2, call: &ChromicPDF.ResetTarget.blank/2, await: &ChromicPDF.ResetTarget.blanked/2, await: &ChromicPDF.ResetTarget.fsl_after_blank/2, output: &ChromicPDF.SpawnSession.output/1], state: %{__protocol__: ChromicPDF.SpawnSession, chrome_args: "--disable-dev-shm-usage", discard_stderr: false, ignore_certificate_errors: false, no_sandbox: true, offline: true, on_demand: false}}, timeout: 5000})
[...]

PR #220, opened 2023-01-19
[#219] Handle DOWN in terminate_worker/3

maltoe:
This patch handles the `DOWN` reason in the `terminate_worker/3` callback, which we previously thought "would not happen", but a user reported it as "does happen indeed". Not 100% sure how or why it happens, but we don't expect anything to break if we handle it.

Fixes [#219](https://github.com/bitcrowd/chromic_pdf/issues/219)

### DIFF
--- a/lib/chromic_pdf/pdf/browser/session_pool.ex
+++ b/lib/chromic_pdf/pdf/browser/session_pool.ex
@@ -188,8 +188,9 @@ defmodule ChromicPDF.Browser.SessionPool do
   # Reasons we want to gracefully clean up the target in the Browser:
   # - max_session_uses_reached, our own mechanism for keeping memory bloat in check
   # - error, when an exception is raised in the Channel
+  # - DOWN, client link is broken (reported by a user as "this happens")
   def terminate_worker(reason, worker_state, pool_state)
-      when reason in [:max_session_uses_reached, :error] do
+      when reason in [:max_session_uses_reached, :error, :DOWN] do
     Task.async(fn ->
       protocol = CloseTarget.new(targetId: worker_state.session.target_id)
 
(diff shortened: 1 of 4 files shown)


lib/chromic_pdf/pdf/browser/session_pool.ex
```diff
@@ -188,8 +188,9 @@ defmodule ChromicPDF.Browser.SessionPool do
   # Reasons we want to gracefully clean up the target in the Browser:
   # - max_session_uses_reached, our own mechanism for keeping memory bloat in check
   # - error, when an exception is raised in the Channel
+  # - DOWN, client link is broken (reported by a user as "this happens")
   def terminate_worker(reason, worker_state, pool_state)
-      when reason in [:max_session_uses_reached, :error] do
```
andreasknoepfle, line 192: I guess we can't easily test this, right?

andreasknoepfle (approved):

lib/chromic_pdf/pdf/browser/session_pool.ex
maltoe: Not really, no. One way would be

1. Spawn a task in the test, run `print_to_pdf` with a long PDF
2. Shoot this task :gun: 
3. Assert that the session is restarted (see `assert_target_respawn` in `SessionPoolTest`)

not sure if I have the motivation to try it, though

andreasknoepfle: Was just asking. I think it is all fine without that 🤷

maltoe: tested now

maltoe:
@andreasknoepfle Made it output a warning through `Logger.warn/1` and added a test. The test may be a bit brittle, but it works :shrug:

andreasknoepfle (approved):

PR #222, opened 2023-01-27
compability with Elixir versions below  1.12 

bclaud:
Hey there,

Since Elixir version requirement isn't specified, I'd love to make the latest version of the lib compatible with versions below 1.12, where Kernel.then was introduced.

Removing it was enough to make the lib compile using at least Elixir 1.11, haven't tried another versions.

Consider 1.10 almost deprecated https://hexdocs.pm/elixir/compatibility-and-deprecations.html

By the way, first PR in Open Source project, so let me know if something isn't right :)


maltoe:
Hey @bclaud 

in that case, congrats on your first OSS contribution :tada:

maltoe (approved):

Issue #233, opened 2023-02-22
on_demand implementation sends exit signal to caller


dvic:
With the current `on_demand` implementation, the caller will receive an `{:EXIT, _supervisor_pid, :normal}` signal, because the supervision tree is started on demand with `Supervisor.start_link` and stopped right after the request with `Supervisor.stop`.

Is this something that's intended?

Background context:

I ran into issues with commanded after this PR https://github.com/commanded/commanded/pull/512. An event handler was calling ChromicPDF and was being shut down because of the above.

maltoe:
Hey @dvic 

could you give me a minimal example to reproduce this?

```elixir
iex(1)> ChromicPDF.start_link(on_demand: true)
{:ok, #PID<0.314.0>}
iex(2)> {:ok, _} = ChromicPDF.print_to_pdf({:html, "test"}); nil
nil
iex(3)> flush
:ok
```

dvic:
The trick is to trap exits:

```elixir
iex(1)> Mix.install([{:chromic_pdf, "~> 1.7.0"}])
:ok
iex(2)> Process.flag(:trap_exit, true)
false
iex(3)> ChromicPDF.start_link(on_demand: true)
{:ok, #PID<0.207.0>}
iex(4)> {:ok, _} = ChromicPDF.print_to_pdf({:html, "test"}); nil
nil
iex(5)> flush
{:EXIT, #PID<0.209.0>, :normal}
:ok
iex(6)>
```

dvic:
Maybe I'm mistaken here and it's completely fine like this because you're supposed to ignore exits with reason `:normal`? It's just that I was not expecting this situation to be triggered by a call to `ChromicPDF.print_to_pdf`.

maltoe:
Oh you're trapping the exit signal. Yes in that case this makes sense. The supervisor that is spawned is linked to your client process and is terminated again with reason `:normal`, so this signal is normally ignored by your process. That's not ChromicPDF, this does the same:

```elixir
iex(1)> Process.flag(:trap_exit, true)
false
iex(2)> spawn_link(fn -> :ok end)
#PID<0.113.0>
iex(3)> flush
{:EXIT, #PID<0.113.0>, :normal}
:ok
```

maltoe:
[...]

PR #234, opened 2023-02-22
Use DynamicSupervisor for on_demand mode.

maltoe:
relates to #233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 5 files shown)


maltoe:
@andreasknoepfle Please :eyes: this only when you have a lot of time. It's not urgent.

In fact, I'm second guessing myself:

```elixir
id = spawn(fn ->
  try do
    :timer.sleep(1000)
  after
    IO.puts("never executes")
  end
end)

Process.exit(pid, :kill)
:timer.sleep(1000)
```

When the client process exits, the `after` is never executed and hence the DynamicSupervisor child will not be terminated :scream: If anything, `OnDemand` should be a proper GenServer and make sure of the termination?

maltoe:
Yeah, nevermind this.

PR #235, opened 2023-02-23
Simplify on_demand mode.

maltoe:


lib/chromic_pdf/supervisor.ex
```diff
@@ -26,69 +26,38 @@ defmodule ChromicPDF.Supervisor do
...
-      else
-        :supervisor
-      end
-
+  def child_spec(config) do
     %{
-      id: chromic,
-      start: {chromic, :start_link, [config]},
-      type: type
+      id: Keyword.fetch!(config, :name),
```
maltoe, line 40: drive-by: also make the `id` dependent on the name to avoid conflicts when dynamic names are used

lib/chromic_pdf/supervisor.ex
```diff
@@ -125,39 +84,38 @@ defmodule ChromicPDF.Supervisor do
...
-      {:found, fun.(pid)}
+  defp with_browser(supervisor, fun) do
+    if agent = find_supervisor_child(supervisor, Agent) do
+      with_on_demand_browser(agent, fun)
     else
-      :not_found
+      supervisor
+      |> find_supervisor_child!(Browser)
+      |> fun.()
     end
```
andreasknoepfle, line 103: Wouldn't it make more sense to swap these around and rather find the `Browser` first, since that seems the more default use case?

andreasknoepfle (approved):

Issue #237, opened 2023-03-23
Timeouts on prod


silverdr:
Out of a sudden (cannot really be sure when it started) I began receiving timeouts on production like the one below. Things that _might_ have changed are

- phoenix version 1.6.x => 1.7.2
- sentry installed according to https://docs.sentry.io/platforms/elixir/
- maybe chromium version

but I tried rolling back all the above changes, removing all Sentry related lines, downgreading chromium, downgrading Phoenix, downgrading ChromicPDF, extending timeout, ...

Also the setup with 1.7.2 and latest chromium works on dev environment. Does anything from the log below ring any bells what might be causing the timeouts? Any clues what else to try?

```
Mar 23 22:43:10 cld bash[12537]: ** (exit) an exception was raised:
Mar 23 22:43:10 cld bash[12537]:     ** (ChromicPDF.Browser.ExecutionError) Timeout in Channel.run_protocol/3!
Mar 23 22:43:10 cld bash[12537]: The underlying GenServer.call/3 exited with a timeout. This happens when the browser was
Mar 23 22:43:10 cld bash[12537]: not able to complete the current operation (= PDF print job) within the configured
Mar 23 22:43:10 cld bash[12537]: 30000 milliseconds.
Mar 23 22:43:10 cld bash[12537]: If you are printing large PDFs and expect long processing times, please consult the
Mar 23 22:43:10 cld bash[12537]: documentation for the `timeout` option of the session pool.
Mar 23 22:43:10 cld bash[12537]: If you are *not* printing large PDFs but your print jobs still time out, this is likely a
Mar 23 22:43:10 cld bash[12537]: bug in ChromicPDF. Please open an issue on the issue tracker.
Mar 23 22:43:10 cld bash[12537]: ---
Mar 23 22:43:10 cld bash[12537]: Current protocol:
Mar 23 22:43:10 cld bash[12537]: %ChromicPDF.Protocol{
Mar 23 22:43:10 cld bash[12537]:   steps: [
Mar 23 22:43:10 cld bash[12537]:     await: &ChromicPDF.Navigate.page_load_event/2,
Mar 23 22:43:10 cld bash[12537]:     call: &ChromicPDF.PrintToPDF.print_to_pdf/2,
Mar 23 22:43:10 cld bash[12537]:     await: &ChromicPDF.PrintToPDF.printed/2,
Mar 23 22:43:10 cld bash[12537]:     call: &ChromicPDF.ResetTarget.reset_history/2,
[...]

PR #238, opened 2023-03-25
Add :raise_on_unhandled_runtime_exceptions option

maltoe:
This patch adds the `:unhandled_runtime_exceptions` option that allows to configure sessions to raise a ChromeError when Chrome sends us a `Runtime.exceptionThrown` notification.

Relates to #237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Added
+
+- Add `unhandled_runtime_exceptions` option that allows to raise on unhandled JavaScript exceptions in templates. Default to warn about them in log.
+
 ### Changed
 
 - Simplified `on_demand` mode. Instead of spawning the entire supervision tree temporarily, we now only spawn the `Browser` process. Also instead of giving the normal `ChromicPDF` name to the temporary supervisor, which effectively prevented concurrent access to ChromicPDF when configured with `on_demand: true`, the temporary `Browser` process remains anonymous.
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -287,6 +287,19 @@ defmodule ChromicPDF do
 
   * `[:chromic_pdf, :join_pdfs, :start | :stop | exception]`
 
+  ## Further options
+
+  ### Debugging unhandled runtime exceptions
+
+  By default, runtime exceptions thrown in JavaScript execution contexts are logged. You may
+  choose to instead convert them into an Elixir exception by passing the following option:
+
+      defp chromic_pdf_opts do
+        [unhandled_runtime_exceptions: :raise]
+      end
+
+  Alternatively, you can pass `:ignore` to silence the log statement.
+
   ## On Accessibility / PDF/UA
 
   Since its version 85, Chrome generates "Tagged PDF" files [by
--- a/lib/chromic_pdf/pdf/protocols/spawn_session.ex
+++ b/lib/chromic_pdf/pdf/protocols/spawn_session.ex
@@ -43,6 +43,12 @@ defmodule ChromicPDF.SpawnSession do
       # Intentionally not awaiting the response to speed up session spawning.
     end
 
+    # Enable Runtime (JS) events, mostly for Runtime.exceptionThrown.
+    # Again, no need to wait for result.
+    if_option {:unhandled_runtime_exceptions, [:log, :raise]} do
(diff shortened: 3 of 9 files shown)


lib/chromic_pdf/supervisor.ex
```diff
@@ -244,6 +244,7 @@ defmodule ChromicPDF.Supervisor do
       @type global_option ::
               {:offline, boolean()}
               | {:disable_scripts, boolean()}
+              | {:raise_on_unhandled_runtime_exceptions, boolean()}
```
andreasknoepfle, line 247: to me, still this would be the better default, adding `ignore_js_errors: true` as an option instead, to explicitly ignore things...

maltoe: But assuming we may have users with templates with unhandled exceptions out there (printing may still work just fine), this would be a breaking change... idk

I'd instead consider warning about them unconditionally (so regardless of this option), wdyt?

lib/chromic_pdf/pdf/protocols/spawn_session.ex
```diff
@@ -43,6 +43,12 @@ defmodule ChromicPDF.SpawnSession do
       # Intentionally not awaiting the response to speed up session spawning.
     end
 
+    if_option {:raise_on_unhandled_runtime_exceptions, true} do
+      call(:runtime_enable, "Runtime.enable", [], %{})
```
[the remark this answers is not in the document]
maltoe: it's the same as `Page.enable`, only for `Runtime` (= JS) events, more or less :) I'll add a comment

lib/chromic_pdf/supervisor.ex
andreasknoepfle: Warning sounds good. Maybe we can still have the option to silence those warnings if they appear explicitly (because JS...)

andreasknoepfle: I just guess not ignoring JS issues by default completely without output will save us some issue work in the long run.

maltoe: ^^ hence I stopped implementing it when I realized that. Solutions would be:

1) somehow store the global options and pass them down to the print job
2) make this option a print job option (i.e. per `print_to_pdf` call), although IMO it feels more like a global option

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Added
+
+- Add `raise_on_unhandled_runtime_exceptions` *global* option that allows to catch JavaScript errors (in templates).
```
maltoe, line 5: `s/raise_on_//g`

maltoe:
@andreasknoepfle I've changed this PR around a lot now. New behaviour is "log by default, ignore or raise as option". For this I needed to change a couple things: 1) In the `SessionPool`, we need to keep the global args around and merge them into the job args, 2) the protocol matcher functions need a way to say "match, but don't remove me" as otherwise the log statement would happen multiple times.

Wdyt?

lib/chromic_pdf.ex
```diff
@@ -287,6 +287,17 @@ defmodule ChromicPDF do
 
   * `[:chromic_pdf, :join_pdfs, :start | :stop | exception]`
 
+  ## Futher options
+
+  ### Debugging unhandled runtime exceptions
+
+  By default, runtime exceptions thrown in JavaScript execution contexts are ignored. You may
```
andreasknoepfle, line 294:
```suggestion
  By default, runtime exceptions thrown in JavaScript execution contexts are logged. You may
```

lib/chromic_pdf.ex
```diff
@@ -287,6 +287,17 @@ defmodule ChromicPDF do
...
+  choose to instead convert them into an Elixir exception by passing the following option:
+
+      defp chromic_pdf_opts do
+        [unhandled_runtime_exceptions: :raise]
+      end
+
   ## On Accessibility / PDF/UA
```
andreasknoepfle, line 301: maybe add another section on how to get rid of JS warnigns

lib/chromic_pdf.ex
```diff
@@ -287,6 +287,17 @@ defmodule ChromicPDF do
 
   * `[:chromic_pdf, :join_pdfs, :start | :stop | exception]`
 
+  ## Futher options
```
andreasknoepfle, line 290:
```suggestion
  ## Further options
```

PR #244, opened 2023-04-28
Gracefully handle errors returned by chrome

tonnenpinguin:
Hi!

While playing around with this I noticed that if I pass some arguably invalid options like in the following example, chromic ignores the error sent by Chrome and instead waits for the timeout and then fails with a generic message.

```
ChromicPDF.print_to_pdf({:html, "<h1>Hello,world!</h1>"}, print_to_pdf: %{pageRanges: "2-3"})
```

In the logs you can see that Chrome immediately tells us that the page range argument was faulty.
```
18:12:15.260 [debug] [ChromicPDF] msg out: %{"id" => 23, "method" => "Page.printToPDF", "params" => %{"footerTemplate" => "", "headerTemplate" => "", "pageRanges" => "2-3"}, "sessionId" => "CE865B72566BD21A8272A3F7EEF30EBE"}
18:12:15.262 [debug] [ChromicPDF] msg in: %{"error" => %{"code" => -32000, "message" => "Page range exceeds page count"}, "id" => 23, "sessionId" => "CE865B72566BD21A8272A3F7EEF30EBE"}
18:12:15.262 [debug] [ChromicPDF] msg in: %{"method" => "Page.frameResized", "params" => %{}, "sessionId" => "CE865B72566BD21A8272A3F7EEF30EBE"}
[... about 5 seconds later the error message appears]
** (ChromicPDF.Browser.ExecutionError) Timeout in Channel.run_protocol/3!

The underlying GenServer.call/3 exited with a timeout. This happens when the browser was
not able to complete the current operation (= PDF print job) within the configured
5000 milliseconds.

If you are printing large PDFs and expect long processing times, please consult the
documentation for the `timeout` option of the session pool.

If you are *not* printing large PDFs but your print jobs still time out, this is likely a
bug in ChromicPDF. Please open an issue on the issue tracker.

---

Current protocol:

%ChromicPDF.Protocol{
  steps: [
    await: &ChromicPDF.PrintToPDF.printed/2,
    call: &ChromicPDF.ResetTarget.reset_history/2,
    await: &ChromicPDF.ResetTarget.history_reset/2,
    call: &ChromicPDF.ResetTarget.blank/2,
    await: &ChromicPDF.ResetTarget.blanked/2,
    await: &ChromicPDF.ResetTarget.fsl_after_blank/2,
    output: &ChromicPDF.PrintToPDF.output/1
  ],
  state: %{
    :__protocol__ => ChromicPDF.PrintToPDF,
    :capture_screenshot => %{},
    :html => "[FILTERED]",
    :ignore_certificate_errors => false,
    :last_call_id => 23,
    :name => "[FILTERED]",
    :offline => false,
    :print_to_pdf => %{
      "footerTemplate" => "[FILTERED]",
      "headerTemplate" => "[FILTERED]",
      "pageRanges" => "2-3"
    },
    :source_type => :html,
    :unhandled_runtime_exceptions => "[FILTERED]",
    "frameId" => "43E1D2D96EBAEDF4BB82A95B74DE397A",
    "sessionId" => "CE865B72566BD21A8272A3F7EEF30EBE"
  }
}

    (chromic_pdf 1.8.1) lib/chromic_pdf/pdf/browser/channel.ex:23: ChromicPDF.Browser.Channel.run_protocol/3
    (chromic_pdf 1.8.1) lib/chromic_pdf/pdf/browser/session_pool.ex:115: ChromicPDF.Browser.SessionPool.do_run_protocol/4
    (nimble_pool 1.0.0) lib/nimble_pool.ex:349: NimblePool.checkout!/4
    (chromic_pdf 1.8.1) lib/chromic_pdf/pdf/browser/session_pool.ex:55: ChromicPDF.Browser.SessionPool.run_protocol/3
    (chromic_pdf 1.8.1) lib/chromic_pdf/api.ex:67: anonymous fn/3 in ChromicPDF.API.chrome_export/4
    (chromic_pdf 1.8.1) lib/chromic_pdf/api/telemetry.ex:10: anonymous fn/2 in ChromicPDF.Telemetry.with_telemetry/3
    (telemetry 1.1.0) /Users/bernhardhackl/src/chromic_pdf/deps/telemetry/src/telemetry.erl:320: :telemetry.span/3
    iex:4: (file)
18:12:20.252 [debug] [ChromicPDF] msg out: %{"id" => 24, "method" => "Target.closeTarget", "params" => %{targetId: "43E1D2D96EBAEDF4BB82A95B74DE397A"}}
```

This PR fixes this and instead of waiting for a timeout handles errors sent via rpc right away.
There's one failing test, but as far as I can tell we should rewrite the test as it basically tests for the behaviour that I fixed here.

### DIFF
--- a/lib/chromic_pdf/pdf/connection/json_rpc.ex
+++ b/lib/chromic_pdf/pdf/connection/json_rpc.ex
@@ -68,10 +68,12 @@ defmodule ChromicPDF.Connection.JsonRPC do
     Map.has_key?(msg, "method") && msg["method"] == method
   end
 
-  def is_error_message?(msg) do
+  @spec is_error?(message()) :: boolean()
+  def is_error?(msg) do
     Map.has_key?(msg, "error")
   end
 
+  @spec extract_error(message()) :: String.t()
   def extract_error(%{"error" => %{"message" => message, "code" => code}}) do
     "#{message} (Code #{code})"
   end
--- a/lib/chromic_pdf/pdf/connection/json_rpc.ex
+++ b/lib/chromic_pdf/pdf/connection/json_rpc.ex
@@ -74,7 +74,7 @@ defmodule ChromicPDF.Connection.JsonRPC do
   end
 
   @spec extract_error(message()) :: String.t()
-  def extract_error(%{"error" => %{"message" => message, "code" => code}}) do
-    "#{message} (Code #{code})"
+  def extract_error(%{"error" => error}) do
+    "#{error["message"]} (Code #{error["code"]})"
   end
 end
--- a/lib/chromic_pdf/pdf/connection/json_rpc.ex
+++ b/lib/chromic_pdf/pdf/connection/json_rpc.ex
@@ -59,7 +59,8 @@ defmodule ChromicPDF.Connection.JsonRPC do
 
   @spec is_response?(message(), call_id()) :: boolean()
   def is_response?(msg, call_id) do
-    Map.has_key?(msg, "result") && msg["id"] == call_id
+    (Map.has_key?(msg, "result") or Map.has_key?(msg, "error")) and
+      msg["id"] == call_id
   end
 
   @spec is_notification?(message(), method()) :: boolean()
@@ -69,5 +70,9 @@ defmodule ChromicPDF.Connection.JsonRPC do
 
   def is_error_message?(msg) do
     Map.has_key?(msg, "error")
+  end
+
+  def extract_error(%{"error" => %{"message" => message, "code" => code}}) do
+    "#{message} (Code #{code})"
   end
 end
--- a/lib/chromic_pdf/pdf/connection/json_rpc.ex
+++ b/lib/chromic_pdf/pdf/connection/json_rpc.ex
@@ -66,4 +66,8 @@ defmodule ChromicPDF.Connection.JsonRPC do
   def is_notification?(msg, method) do
     Map.has_key?(msg, "method") && msg["method"] == method
   end
+
+  def is_error_message?(msg) do
+    Map.has_key?(msg, "error")
+  end
 end
--- a/test/integration/pdf_generation_test.exs
+++ b/test/integration/pdf_generation_test.exs
@@ -415,8 +415,8 @@ defmodule ChromicPDF.PDFGenerationTest do
       ]
 
       assert capture_log(fn ->
-               assert_raise ChromicPDF.Browser.ExecutionError,
-                            ~r/Timeout in Channel.run_protocol/,
+               assert_raise ChromicPDF.ChromeError,
+                            ~r/Chrome returned an error/,
                             fn ->
                               ChromicPDF.print_to_pdf({:html, ""}, params)
                             end
--- a/test/integration/pdf_generation_test.exs
+++ b/test/integration/pdf_generation_test.exs
@@ -323,6 +323,19 @@ defmodule ChromicPDF.PDFGenerationTest do
     end
   end
 
+  describe "handle errors returned by chrome" do
+    setup do
+      start_supervised!(ChromicPDF)
+      :ok
+    end
+
+    test "raise nicely formatted errors" do
+      assert_raise ChromicPDF.ChromeError, ~r/Chrome returned an error/, fn ->
+        print_to_pdf({:html, test_html()}, print_to_pdf: %{pageRanges: "2-3"})
+      end
+    end
+  end
+
   describe "a cookie can be set when printing" do
     @cookie %{
       name: "foo",
--- a/test/integration/pdf_generation_test.exs
+++ b/test/integration/pdf_generation_test.exs
@@ -330,7 +330,7 @@ defmodule ChromicPDF.PDFGenerationTest do
     end
 
     test "raise nicely formatted errors" do
-      assert_raise ChromicPDF.ChromeError, ~r/Chrome returned an error/, fn ->
+      assert_raise ChromicPDF.ChromeError, ~r/Page range exceeds page count/, fn ->
         print_to_pdf({:html, test_html()}, print_to_pdf: %{pageRanges: "2-3"})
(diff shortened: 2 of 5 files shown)


test/integration/pdf_generation_test.exs
```diff
@@ -402,8 +415,8 @@ defmodule ChromicPDF.PDFGenerationTest do
       ]
 
       assert capture_log(fn ->
-               assert_raise ChromicPDF.Browser.ExecutionError,
-                            ~r/Timeout in Channel.run_protocol/,
+               assert_raise ChromicPDF.ChromeError,
+                            ~r/Chrome returned an error/,
```
tonnenpinguin, line 419: I'm not sure about this one. Is this also fine?

maltoe:
Hey @tonnenpinguin,

thanks for the contribution! Looks good and is clearly a way to do this, though I'd prefer a different solution. Let me know what you think:

- Side-note on `intercept_exception_thrown`: I've recently added this handler before every `await_notification` function to look for a specific error _notification_ which may occur at any time. This could eventually be made more generic to intercept other failed notifications, though we do receive a bunch of them that from ChromicPDF's POV are fine - as in, they don't matter for PDF printing. So for now this is specifically for the JS exceptions.
- The error you're after is a _response error_, i.e. it carries the `id` field matching it to a call we sent previously. This we can better handle inside the normal `await_response` handler. In fact, there is already some code that does handle errors in responses. See the mechanism https://github.com/bitcrowd/chromic_pdf/blob/4df426e7f14ce26edb01bab2777d6ff8ace80101/lib/chromic_pdf/pdf/protocol_macros.ex#L193-L197 in `Protocol`, and [this use](https://github.com/bitcrowd/chromic_pdf/blob/4df426e7f14ce26edb01bab2777d6ff8ace80101/lib/chromic_pdf/pdf/protocols/navigate.ex#L26-L32) of it in the `Navigate` protocol (where for some reason we only treat messages with `errorText != nil` as an error).

So, I'd suggest to do the following:

- extend `JsonRPC.is_response?` to also return true for error response (= `id` & `error`) key
- add a default implementation for the `await_response` callback, in the lines I linked in `ProtocolMacros`. In other words, if the protocol does not export a callback, use a default one. You can probably get away with a `cond` instead of the `if` and make `JsonRPC.is_error?` the second condition.
- error handling should then already be implemented, i.e. you can probably drop your `error_returned` extras in `ChromeError`. Also the test you commented on would not be affected.

Benefits of this approach

- It leaves us the option to ignore errors, or only treat them as errors under certain conditions. 
- We don't run this code for every message we receive. Now, this also means that we won't see (yet unknown) errors for calls that don't define a `await_response`, but that's a trade-off between performance and developer experience where I'd lean towards performance  (= reducing needlessly executed code). Just consider that after you've dealt with your "page range exceeds page count" error, it is unlikely to happen ever again. At least in our experience, we usually set up a template and once it works, it works.

test/integration/pdf_generation_test.exs
maltoe: Yes, changing the error is fine here, the test should assert on the log statement.

maltoe:
Moin moin nach Hamburg im Übrigen

tonnenpinguin:
Thanks for the detailed feedback, that helps a lot when playing around with a new codebase :)

I applied the suggested changes, let me know what you think! :)

A couple random things I noticed
* ChromicPDF.Browser.Channel.warn_on_inspector_crash!/1 has an [trailing bang](https://hexdocs.pm/elixir/1.12/naming-conventions.html#trailing-bang-foo) even though there's no assertions in the function body 
* `ChromicPDF.Connection.JsonRPC.is_response?/2` et al have a [is_prefix](https://hexdocs.pm/elixir/1.12/naming-conventions.html#is_-prefix-is_foo) even though they are not valid in guard clauses

lib/chromic_pdf/pdf/connection/json_rpc.ex
```diff
@@ -59,11 +59,22 @@ defmodule ChromicPDF.Connection.JsonRPC do
...
   end
+
+  @spec is_error?(message()) :: boolean()
+  def is_error?(msg) do
+    Map.has_key?(msg, "error")
+  end
+
+  @spec extract_error(message()) :: String.t()
+  def extract_error(%{"error" => error}) do
+    "#{error["message"]} (Code #{error["code"]})"
```
tonnenpinguin, line 78: I wasn't sure about this one.
If I just pass the error Map we'll need some extra handling in ChromeError as the way it's coded now it only supports binaries.

maltoe: It's great like this :+1: IMO handling it here is better than in `ChromeError`, which doesn't know about JSONRPC

test/integration/pdf_generation_test.exs
```diff
@@ -323,6 +323,19 @@ defmodule ChromicPDF.PDFGenerationTest do
     end
   end
 
+  describe "handle errors returned by chrome" do
```
maltoe, line 326:
```suggestion
  describe "generic handling of protocol response errors" do
```

test/integration/pdf_generation_test.exs
```diff
@@ -323,6 +323,19 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+    setup do
+      start_supervised!(ChromicPDF)
+      :ok
+    end
+
+    test "raise nicely formatted errors" do
```
maltoe, line 332:
```suggestion
    test "response errors raise nicely formatted errors" do
```

maltoe (approved):
Nice work! Thanks again :100: 

Could you add an entry to the `CHANGELOG.md`? Then we're good to :ship:

maltoe:
> A couple random things I noticed

Feel free to fix these, if you like.

maltoe:
merged to `main` manually

Issue #224, opened 2023-01-28
Flaky test assertion on log output


maltoe:
![image](https://user-images.githubusercontent.com/96114/215279982-eecc355f-6749-416c-b25c-d9826c13b579.png)

tonnenpinguin:
I also noticed this happening on my local machine, though it doesn't happen when you run the failing test in isolation 🤷

maltoe:
Haven't seen it once on my local machine so far :shrug: It's probably a timing issue, as the `Inspector.targetCrashed` message is a notification, possibly arriving after the `printToPDF` error response.

maltoe:
Test disabled for now, remove the `@tag :skip` to see it fail on CI again

PR #246, opened 2023-05-01
[#224] Flaky test assertion on log output

maltoe:
[[#224]](https://github.com/bitcrowd/chromic_pdf/issues/224)


maltoe:
observations:

1. this doesn't help much
2. instead of the empty `capture_log` result I now get timeouts

PR #250, opened 2023-05-22
Update Alpine 3.17 test image

maltoe:


gmile:
Out of curiosity, why not update to Alpine 3.18?

maltoe:
> Out of curiosity, why not update to Alpine 3.18?

I think at the time I started this PR, the relevant [hexpm/elixir package](https://hub.docker.com/layers/hexpm/elixir/1.14.5-erlang-25.3.1-alpine-3.18.0/images/sha256-f6041ad1d08c30e3b5b1237930d440d84228de0f2860396d07f81d3f7094b8e3?context=explore) wasn't available for 3.18. No real reason. But I also don't expect any incompatibility with Alpine 3.18, if you intend to use that.

gmile:
@maltoe makes sense :+1:

PR #256, opened 2023-06-09
Add :full_page option to capture_screenshot/2

maltoe:
See https://github.com/puppeteer/puppeteer/blob/9c00c2c57e8b698c266a9b463c4601ec730d033f/packages/puppeteer-core/src/common/Page.ts#L1362-L1389 for reference


hannahbit (approved):
Nice option! LGTM 🚀

PR #263, opened 2023-06-20
Replace deprecated `Logger.warn/1` on Elixir 1.15.0

aptinio:
with `Logger.warning/1` (available since Elixir 1.11.0)


maltoe:
Thank you :purple_heart:

PR #265, opened 2023-07-12
Add :console_api_calls global option.

maltoe:
This patch adds the `:console_api_calls` global option, which works the same as the `:unhandled_runtime_exception` option in that it can be set to `:ignore`, `:log`, or `:raise`. By default, console API calls are ignored.

### DIFF
--- a/lib/chromic_pdf/pdf/protocol_macros.ex
+++ b/lib/chromic_pdf/pdf/protocol_macros.ex
@@ -156,12 +156,9 @@ defmodule ChromicPDF.ProtocolMacros do
       @steps {:await, unquote(name), unquote(length(args))}
 
       def unquote(fundef) do
-        case intercept_exception_thrown(unquote_splicing(args)) do
-          :no_match ->
-            unquote(block)
-
-          error ->
-            error
+        with :no_match <- intercept_exception_thrown(unquote_splicing(args)),
+             :no_match <- intercept_console_api_called(unquote_splicing(args)) do
+          unquote(block)
         end
       end
     end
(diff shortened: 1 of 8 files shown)


lib/chromic_pdf/pdf/connection/inet.ex
```diff
@@ -58,7 +58,7 @@ if Code.ensure_loaded?(WebSockex) do
       :inets.start()
 
       url = String.to_charlist("http://#{host}:#{port}/json/version")
-      headers = [{'accept', 'application/json'}]
+      headers = [{~c"accept", ~c"application/json"}]
```
maltoe, line 61: :thinking: new elixir version makes formatter do funny things

test/integration/support/get_targets.ex
```diff
@@ -15,7 +15,9 @@ defmodule ChromicPDF.GetTargets do
   def run do
     {:ok, target_infos} =
       ChromicPDF.Supervisor.with_services(ChromicPDF, fn services ->
-        ChromicPDF.Browser.run_protocol(services.browser, __MODULE__, skip_session_use_count: true)
+        ChromicPDF.Browser.run_protocol(services.browser, __MODULE__,
+          skip_session_use_count: true
+        )
```
maltoe, line 20: same here

lib/chromic_pdf/pdf/connection/inet.ex
maltoe: resolved by pinning the local elixir version to 1.14 as that's the one the `lint` job uses

lib/chromic_pdf/pdf/protocol_macros.ex
```diff
@@ -235,22 +232,50 @@ defmodule ChromicPDF.ProtocolMacros do
     with true <- JsonRPC.notification?(msg, "Runtime.exceptionThrown"),
          true <- state["sessionId"] == msg["sessionId"] do
       exception = get_in!(msg, ["params", "exceptionDetails"])
+      description = get_in!(exception, ["exception", "description"])
 
       case Map.get(state, :unhandled_runtime_exceptions, :log) do
         :ignore ->
-          :no_match
+          {:match, :keep, state}
```
hannahbit, line 239: Why does this no longer return `:no_match`?  What is the purpose of this change?

I am confused, as this will lead in the with clause of line 159 to not match anymore and thus return `{:match, :keep, state}` instead of running the block. Or am I missing something?

hannahbit (approved):
Looks good to me ✨ 

Left just some tiny remarks.

lib/chromic_pdf/pdf/protocol_macros.ex
maltoe: Yes, I reconsidered this code while working on the (very similar) logic for `console_api_calls`: Figured that even though we ignore the exception, it doesn't really make sense to match it against the actual awaited message pattern. Or does it? Couldn't fathom a scenario where I'd wait for an exception to be thrown before continueing in a protocol.

test/integration/pdf_generation_test.exs
```diff
@@ -323,6 +323,53 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+    end
+  end
+
+  describe "Console API calls with console_api_calls: :log option" do
+    setup do
+      start_supervised!({ChromicPDF, console_api_calls: :log})
+      :ok
+    end
+
+    test "are logged by default" do
```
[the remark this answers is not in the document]
maltoe: idk, with `by default` it's consistent with the block above for `unhandled_runtime_exceptions`, so I think I'd prefer to keep it.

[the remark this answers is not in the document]
hannahbit: Hmm. But logging is the default for `unhandled_runtime_exceptions` whereas it is not the default for `console_api_calls`, no?

PR #273, opened 2023-08-14
Add named session pools

maltoe:
This adds support for multiple named session pools.

I figured that people in the past repeatedly wanted to have different values for our "global" options (e.g. `disable_scripts`) per print job, or rather likely per template. As this goes against my original idea of keeping "initialization work" strictly to the initialization (`SpawnSession`) and keep the amount of work per print job to a minimum, I'm inclined to point people to the existing possibility of launching multiple ChromicPDF instances with different configs (think: print queues). However, launching a separate ChromicPDF supervisor means launching additional OS processes of chromium, which seems wasteful. At this point I realized that we're lacking a middle layer in the session pool, i.e. differently configured sessions. This PR fixes that by introducing **named session pools** that can have pool-specific configs.

## API

```elixir
iex> ChromicPDF.start_link(session_pool: %{noscript: [disable_scripts: true], withscripts: []})
iex> ChromicPDF.print_to_pdf({:html, "test"}, session_pool: :noscript)
```

## Docs

image

image

### DIFF
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -182,7 +182,7 @@ defmodule ChromicPDF do
   ### Multiple session pools
 
   ChromicPDF supports running multiple named session pools to allow varying session configuration.
-  For example, this makes it possible to one pool that is not allowed to execute JavaScript while
+  For example, this makes it possible to have one pool that is not allowed to execute JavaScript while
   others can use JavaScript.
 
       defp chromic_pdf_opts do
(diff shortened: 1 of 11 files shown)


lib/chromic_pdf.ex
```diff
@@ -181,6 +179,28 @@ defmodule ChromicPDF do
         ]
       end
 
+  ### Multiple session pools
+
+  ChromicPDF supports running multiple named session pools to allow varying session configuration.
+  For example, this makes it possible to one pool that is not allowed to execute JavaScript while
```
hannahbit, line 185: I think there is a verb missing. Maybe "have"?

```suggestion
  For example, this makes it possible to have one pool that is not allowed to execute JavaScript while
```

hannahbit (commented):
Seems reasonable to me and is nicely documented ✨ 

Though I feel like I am missing a lot of context to properly review.

PR #278, opened 2023-09-27
[#276] Make checkout timeout configurable

### LINKED ISSUE #276: Make checkout timeout configurable
https://elixirforum.com/t/how-to-handle-max-concurrency-for-chromicpdf/58422/9

maltoe:
This patch adds a global `:checkout_timeout` option to configure the allowed wait-time of an operation in the session pool's queue.

This also deliberately drops the section on `:init_timeout` from the docs as I don't see why anyone would mess with that. Same goes for `:close_timeout` (configurable, but not advertised in the docs).

Fixes #276 

cc @axelclark

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Added
+
+- Configurable pool checkout timeout via new global option `:checkout_timeout`.
+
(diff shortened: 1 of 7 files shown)


axelclark:
I think the updated docs and error message look great!

PR #280, opened 2023-10-07
Add Template.options/1

maltoe:
This patch adds `ChromicPDF.Template.options/1` to fetch the options for the `print_to_pdf/2` call without passing in the content.

Besides, styles generate by `Template.styles/1` are now squished.

### DIFF
--- a/lib/chromic_pdf/template.ex
+++ b/lib/chromic_pdf/template.ex
@@ -30,10 +30,9 @@ defmodule ChromicPDF.Template do
 
   require EEx
 
-  @type content_option ::
-          {:content, iodata()}
-          | {:header, iodata()}
-          | {:footer, iodata()}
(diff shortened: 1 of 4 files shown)


lib/chromic_pdf/template.ex
```diff
@@ -202,6 +193,54 @@ defmodule ChromicPDF.Template do
...
+      print_to_pdf: %{
+        displayHeaderFooter: true,
+        preferCSSPageSize: true,
+        headerTemplate: html_concat(styles, header),
+        footerTemplate: html_concat(styles, footer),
+        paperWidth: width,
+        paperHeight: height
+      }
+    ]
+  end
```
andreasknoepfle, line 242: Maybe it could be nice to add a test for the `options` function to make sure that we can in fact pass that in to `print_to_pdf`.

PR #285, opened 2023-12-08
Add another potential & likely default_executable name

joshamb:
It seems the default .exe name on my windows server is 'chrome.exe', and as I assume that's probably fairly common so i've added it to the default list.


maltoe:
@joshamb Thanks! Does it also work if you put `chrome.exe`?

Issue #290, opened 2023-12-13
Footer is not displayed anymore with recent Chromiums


maltoe:
Footer templates set via `ChromicPDF.Template` are not displayed anymore with recent Chromiums.

`ChromicPDF.Template` sets page dimensions and header/footer dimensions via rules in `@page { ... }` and sets `preferCSSPageSize: true` in the `printToPDF` call. Some time after Chromium 117, footers set via `footerTemplate` have disappeared (margin is there but blank).

## Analysis so far

- Somewhere in between Chromium 117.0.5938.62 and 119.0.6045.159
- setting the page margins via `printToPDF` options (in inches) fixes it
- but even if only `margin-top` is present in `@page`, it breaks again
  - even with `preferCSSPageSize: false`

## Minified example of what is broken

```css
@page {
  size: 210mm 297mm;
  /* or any other variation of specifying margin-top and margin-bottom */
  margin: 40mm 0 40mm;
}
```

See [stylesheet](https://github.com/bitcrowd/chromic_pdf/blob/ed525c2400922cc83ebdd41eaa8907de68cb0c55/lib/chromic_pdf/template.ex#L245-L275) for actual code.

## Things we've tried

- `size` vs. `width/height` in `@page` has no effect
- removing the `zoom` rule from `#header, #footer` has no effect
- keeping a `margin-inline: ...` is ok, but specifying top/bottom in the CSS immediately breaks it

janwillemvd:
We encountered the same here! It was introduced in one of projects on 5 October when we deployed a new release of our software with the latest images of our build OS.

maltoe:
@janwillemvd can you pinpoint the chromium version where it broke? We currently only have 117.0.5938.62 as last known good and 119.0.6045.159 as broken.

maltoe:
script to reproduce this https://gist.github.com/maltoe/104d192430a059f3672cb7a8c7a3b1bf (assumes local chrome is 120 or so and chrome 115 is run via docker)

![image](https://github.com/bitcrowd/chromic_pdf/assets/96114/1fc1b043-9c7c-43db-944e-2b37d0ee14c1)

maltoe:
potential culprits:

- anything on this page after Aug 4 https://github.com/chromium/chromium/commits/main/components/printing

maltoe:
analysis so far:

[...]

PR #291, opened 2023-12-15
[#290] Fix missing footers

maltoe:
Fixes [[#290]](https://github.com/bitcrowd/chromic_pdf/issues/290)

Since Chrome 117 custom footers would not be displayed anymore when configured via `ChromicPDF.Template`. Reason for this was that our stylesheet was prepended to both the content markup as well as the header and footer templates. The `@page` rule contains the page margins which used to be ignored in the header and footer sections, but are not applied and lead to the footer being pushes out of its content box. Fixed by splitting `ChromicPDF.Template.styles/1` into `page_styles/1` and `header_footer_styles/1` and removing the `@page` directive from the header and footer styles.

Besides, this patch removes the `zoom` directive from the header and footer templates for Chrome >= v120, as they have removed the default 4/3 scale-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ## Unreleased
 
+### Fixed
+
+- Recover missing custom footer templates by making sure we do not add any `@page` CSS rule to the header or footer templates in `ChromicPDF.Template.source_and_options/1`. (#290)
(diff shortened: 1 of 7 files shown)


hannahbit (commented):
Looks good!

Just wondering about the new utils functions. Should they have their own tests? Or are they implicitely tested already?

maltoe:
> Looks good!
> 
> Just wondering about the new utils functions. Should they have their own tests? Or are they implicitely tested already?

Added some for `semver_compare/2`

PR #299, opened 2024-01-04
Allow phoenix_html 4.0

idabmat:
Chromic does not seem to rely on any functions removed in v4.0.0 : https://hexdocs.pm/phoenix_html/changelog.html


maltoe:
Thank you! :purple_heart:

PR #304, opened 2024-03-11
fix: correct hint about how to set chrome args

tonnenpinguin:
Took me a while to figure out what this comment was actually referring to, so I hope this helps out somebody else 🙂


maltoe:
Thank you, good catch! :fish:

PR #313, opened 2024-07-22
Enable connecting over ipv6

sasikumar87:
Changes to enable connecting over ipv6

### DIFF
--- a/lib/chromic_pdf/pdf/connection/inet.ex
+++ b/lib/chromic_pdf/pdf/connection/inet.ex
@@ -56,12 +56,13 @@ if Code.ensure_loaded?(WebSockex) do
     defp websocket_debugger_url({host, port}) do
       # Ensure inets app is started. Ignore error if it was already.
       :inets.start()
+      :httpc.set_options(ipfamily: :inet6fb4)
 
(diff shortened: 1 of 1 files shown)


maltoe:
Hey @sasikumar87 

may I ask what do you need this for?

sasikumar87:
Hi @maltoe, I am trying to deploy chrome on a fly.io machine in private mode. Fly machine's ipv4 is always public. To make it private, I must use IPv6.
Ref : https://fly.io/docs/networking/app-services/#private-flycast-apps-2

Though I tested it further and found `websokex` also needs change at `:gen_tcp.connect` to make websockets connect over ipv6. https://github.com/Azolo/websockex/blob/v0.4.3/lib/websockex/conn.ex#L156 

So the changes in this PR alone is not sufficient without websockex changes.

sasikumar87:
I'll close this for now, as websockex also needs change.

walter:
@sasikumar87 I'm not seeing a PR or issue on websockex with the required change for ipv6. Is there something in the works to support it?

sasikumar87:
@walter I just raised an issue in websockex https://github.com/Azolo/websockex/issues/131 .

PR #314, opened 2024-07-23
Add ability to remove default args via config

walter:
When trying to use chromic_pdf in combination with pages that contain WebGL images, there are two default args in `ChromeRunner` that prevent it from working.

* `--headless` (rather than `--headless=new`) 
* `--disable-gpu` which, despite its name, will disable software based webgl rendering in addition to hardware accelerated webgl

Based on my experimentation, simply overriding these in `:chrome_args` is not enough. Once they are declared, they can't be overridden by further switches.

Instead of just taking these out, and maybe break other people's projects, I think adding the ability to remove the args via configuration is a more  useful feature.

### DIFF
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -287,13 +287,16 @@ defmodule ChromicPDF do
   E.g. in default args within chromic_pdf's code, `--disable-gpu` and then
   passing `chrome_args: "--enable-gpu"` won't work.
 
-  The `:conflicting_args` option allows targeted removing of problematic default
-  args so that they don't disrupt your chrome configuration.
+  In this case, use Keyword form of the `:chrome_args` option which
+  allows targeted removing of problematic default args so that they don't
+  disrupt your chrome configuration.
 
       defp chromic_pdf_opts do
-        [conflicting_args: ["--headless", "--disable-gpu"],
-         chrome_args: "--headless=new --angle=swiftshader"]
+        [chrome_args: [append: "--headless=new --angle=swiftshader",
+                       remove: ["--headless", "--disable-gpu"]]]
       end
+
+  _Note: `:append` expects a string whereas `:remove` expects a list._
 
   The `:chrome_executable` option allows to specify a custom Chrome/Chromium executable.
 
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -296,8 +296,6 @@ defmodule ChromicPDF do
                        remove: ["--headless", "--disable-gpu"]]]
       end
 
-  _Note: `:append` expects a string whereas `:remove` expects a list._
-
   The `:chrome_executable` option allows to specify a custom Chrome/Chromium executable.
 
       defp chromic_pdf_opts do
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -278,6 +278,23 @@ defmodule ChromicPDF do
         [chrome_args: "--font-render-hinting=none"]
       end
 
+  In some cases, chromic_pdf's default args may conflict with the ones you would like to add.
+
+  E.g. `--disable-gpu`
+
+  The problem is that you can't simply override them like so:
+
+  E.g. in default args within chromic_pdf's code, `--disable-gpu` and then
+  passing `chrome_args: "--enable-gpu"` won't work.
+
+  The `:conflicting_args` option allows targeted removing of problematic default
+  args so that they don't disrupt your chrome configuration.
+
+      defp chromic_pdf_opts do
+        [conflicting_args: ["--headless", "--disable-gpu"],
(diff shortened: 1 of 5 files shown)


lib/chromic_pdf.ex
```diff
@@ -278,6 +278,23 @@ defmodule ChromicPDF do
...
+  The problem is that you can't simply override them like so:
+
+  E.g. in default args within chromic_pdf's code, `--disable-gpu` and then
+  passing `chrome_args: "--enable-gpu"` won't work.
+
+  The `:conflicting_args` option allows targeted removing of problematic default
+  args so that they don't disrupt your chrome configuration.
+
+      defp chromic_pdf_opts do
+        [conflicting_args: ["--headless", "--disable-gpu"],
```
maltoe, line 294: How about, instead of introducing a new option that doesn't relate to chrome anymore, we extend the values the `chrome_args` option can take? E.g. we could make it a `String.t() | [append: String.t(), remove: String.t()]` or so?

maltoe (changes requested):
Hi @walter,

thanks for the contribution! :purple_heart: Code looks good to me, though I'd prefer if we could extend `chrome_args` instead of adding a new option for this.

Please add yourself to the changelog as well, something like

```
## Unreleased

### Added

- ... (@walter)
```

lib/chromic_pdf.ex
walter: > How about, instead of introducing a new option that doesn't relate to chrome anymore, we extend the values the chrome_args option can take? E.g. we could make it a String.t() | [append: String.t(), remove: String.t()] or so?

Like it! 

Made the suggested changes and pushed with the slight variation that `:remove` takes a list instead of string.

lib/chromic_pdf.ex
```diff
@@ -278,6 +278,26 @@ defmodule ChromicPDF do
...
+  In this case, use Keyword form of the `:chrome_args` option which
+  allows targeted removing of problematic default args so that they don't
+  disrupt your chrome configuration.
+
+      defp chromic_pdf_opts do
+        [chrome_args: [append: "--headless=new --angle=swiftshader",
+                       remove: ["--headless", "--disable-gpu"]]]
+      end
+
+  _Note: `:append` expects a string whereas `:remove` expects a list._
```
maltoe, line 299: Let's be liberal in what the API can accept; a `List.wrap/1` somewhere will allow you to accept strings or lists for either of the options, and then we don't need this note in the docs.

```suggestion
```

lib/chromic_pdf/supervisor.ex
```diff
@@ -226,7 +226,7 @@ defmodule ChromicPDF.Supervisor do
       @type local_chrome_option ::
               {:no_sandbox, boolean()}
               | {:discard_stderr, boolean()}
-              | {:chrome_args, binary()}
+              | {:chrome_args, binary() | map()}
```
maltoe, line 229:
```elixir
# above
@type extended_chrome_args :: [{:append, binary() | [binary()]} | {:remove, binary() | [binary()]}]
```

```suggestion
              | {:chrome_args, binary() | extended_chrome_args()}
```

lib/chromic_pdf/pdf/chrome_runner.ex
```diff
@@ -147,15 +147,37 @@ defmodule ChromicPDF.ChromeRunner do
   # https://github.com/bitcrowd/chromic_pdf/issues/76
   defp args(extra, opts) do
     default_args()
+    |> remove_conflicting_args(opts[:chrome_args])
     |> append_if("--no-sandbox", no_sandbox?(opts))
-    |> append_if(to_string(opts[:chrome_args]), !!opts[:chrome_args])
+    |> append_chrome_args_if(opts[:chrome_args])
```
maltoe, line 152: I think the implementation could use another round of polishing :)

```suggestion
    |> handle_chrome_args(opts[:chrome_args])
```

Maybe go for a unified handler function. 3 Clauses, `nil` if option not set, then a `is_binary` clause for the original behaviour, and a new `is_list` clause for handling the extended args. Extracting options from this list should then be something like:

```elixir
remove = List.wrap(Keyword.get(extended, :remove, []))
```

maltoe (changes requested):
Hi @walter 

Had a few stylistic nitpicks still. If you don't want to continue working on this, I can also merge it in and refactor afterwards. As you prefer.

Thanks again!

walter:
> Hi @walter
> 
> Had a few stylistic nitpicks still. If you don't want to continue working on this, I can also merge it in and refactor afterwards. As you prefer.
> 
> Thanks again!

Thanks for the feedback. Jibes with some thoughts I had myself, but was distracted by other work. I hadn't used `List.wrap` before, nice!

Have a look at latest and let me know if there is anything else.

maltoe:
Thanks again!

maltoe:
Released as v1.16.1

PR #328, opened 2024-08-30
Add new headless mode flag

PJUllrich:
Hey folks,

Chrome introduced a new headless mode with Chrome 112: https://developer.chrome.com/docs/chromium/new-headless

ChromicPDF currently uses the old headless mode because the flag `--headless` defaults to `--headless=old`. The problem is that we can't overwrite the default flag with `--headless=new` because the list of arguments would contain both flags and I'm unsure which flag is ignored.

So, I added a quick option, which enables the mode if set to `true`. Please let me know what you think about this and feel free to ignore this PR of course if you think it's unnecessary :)

### DIFF
--- a/lib/chromic_pdf.ex
+++ b/lib/chromic_pdf.ex
@@ -79,6 +79,16 @@ defmodule ChromicPDF do
   Note that this doesn't prevent other features like the `evaluate` option from working, it
   solely applies to scripts being supplied by the rendered page itself.
 
+  ### New vs Old Headless Mode
+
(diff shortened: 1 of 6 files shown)


PJUllrich:
Ah nevermind! I just found the documentation for adding/removing flags. It actually mentions `--headless=new`, so I'll be using that one. Sorry for the oversight of not finding this earlier.

https://github.com/bitcrowd/chromic_pdf/blob/main/lib/chromic_pdf.ex#L286

PR #330, opened 2024-09-26
Add test for alpine 3.20.2 with chrome 128

andreasknoepfle:


.tool-versions
```diff
@@ -1,3 +1,3 @@
 # locked to the versions we use in the lint CI job
 elixir 1.14.5
-erlang 25.3.1
```
andreasknoepfle, line 3: This version is not working properly on a Mac without a lot of hassle.

.github/workflows/publish-image.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       packages: write
     strategy:
       matrix:
-        dockerfile: ['elixir-1-11', 'debian-buster', 'alpine-3-17-3', 'alpine-3-18-4']
+        dockerfile: ['elixir-1-11', 'debian-buster', 'alpine-3-17-3', 'alpine-3-18-4', 'alpine-3-20-2']
```
andreasknoepfle, line 34: Already ran it once on this branch :)

maltoe (commented):
:shrug: Looks good, but still fails on my machine (ie. I've built this locally and ran the tests)

andreasknoepfle:
> 🤷 Looks good, but still fails on my machine (ie. I've built this locally and ran the tests)

Can we try to reproduce this on a CI run somehow?

maltoe:
> > 🤷 Looks good, but still fails on my machine (ie. I've built this locally and ran the tests)
> 
> Can we try to reproduce this on a CI run somehow?

Not sure how, I mean we already run it inside this very container, on comparable x86-64 hardware on Linux. Maybe kernel version or some config issue? Not sure :shrug: 

Should we still merge this?

maltoe:
@andreasknoepfle Chrome is at v143 already, should we update this PR and see if it works again? Then we could just warn about some incompatible versions in the README and be done with it.

andreasknoepfle:
sure. If I remember well you could get it to fail on your system, no?

Issue #327, opened 2024-08-27
Recent versions of chromium do not support --remote-debugging-address


MaeIsBad:
Recent chrome versions only listen on `127.0.0.1`(https://issues.chromium.org/issues/41487252)

In order to access chrome from another machine you need to proxy your traffic through something like `socat`.

For container use cases you need to ensure you're running in the same namespace, so using the same k8s pod, or by using the network_mode option in docker-compose

```
version: "3"

services:
  chromic_pdf:
    build:
      context: .
    depends_on:
      - browser
  browser:
   network_mode: "service:chromic_pdf"
   image: chrome
```

In order to avoid running into cyclic dependency issues you might want to use a pause container to create a network namespace, replicating how k8s pods work

maltoe:
@MaeIsBad thank you for the info! Any chance you can wrap this up and add it as another callout or section in the docs [here](https://hexdocs.pm/chromic_pdf/ChromicPDF.html#module-remote-chrome)?

PR #334, opened 2024-12-19
fix: unable to remotely connect to chrome

clemensm:
Chrome requires the "host" header to be set to either an ip address or "localhost", otherwise it will not accept the connection.

I've fixed it by requesting a "localhost" host, and then manually patching the websocket address in the reply from Chrome. The use case here would be a docker container running with [Zenika/alpine-chrome](https://github.com/Zenika/alpine-chrome), and my application in another docker container, with a shared network. Otherwise this might work with fixed IP's as well.

### DIFF
--- a/lib/chromic_pdf/pdf/connection/inet.ex
+++ b/lib/chromic_pdf/pdf/connection/inet.ex
@@ -58,7 +58,9 @@ if Code.ensure_loaded?(WebSockex) do
       :inets.start()
 
       url = String.to_charlist("http://#{host}:#{port}/json/version")
-      headers = [{~c"accept", ~c"application/json"}]
+      # Chrome requires the "host" header to be set to either an ip address or localhost,
+      # otherwise it will not accept the connection
+      headers = [{~c"accept", ~c"application/json"}, {~c"host", ~c"localhost"}]
       http_request_opts = [ssl: [verify: :verify_none]]
 
       case :httpc.request(:get, {url, headers}, http_request_opts, []) do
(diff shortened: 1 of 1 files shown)


maltoe:
Hi @clemensm 

thank you! Is this related to #327? Meaning, Chrome does accept remote connections when the host header is set? Or do you have to run socat in addition to this?

clemensm:
Hi Malte, yeah, I think this is related to #327, at least I was unable to get a Chrome connection without the header, and without setting it to "localhost" or 127.0.0.1.

So I'm running the docker container using:

```
podman container run -d --rm --name chrome --network development zenika/alpine-chrome:latest --no-sandbox --remote-debugging-address=0.0.0.0 --remote-debugging-     port=9222
```

With the remote-debugging-address still set to 0.0.0.0, but without the header no luck in connecting.

And no, I do not need to use socat in this case, just the container as it is in the above command.

maltoe:
:thinking: 

```
--remote-debugging-address=0.0.0.0
```

But this is specifically what @MaeIsBad said in #327 wouldn't work anymore with recent Chromium. Don't get me wrong, I'm not opposed to the change you suggest at all, but I'd like to have a reasonable plan forward for remote chrome users, ideally incl. documentation. 

What chromium version do you get from `zenika/alpine-chrome:latest` currently? Also, I'm assuming the podman network is a bridge network? Could you try isolate the container and connect through a remote address?

clemensm:
Regarding your questions: 

`zenika/alpine-chrome:latest` is `Chrome:124`, the newest tag I can find in their container repositories, so maybe with a newer Chrome this will no longer work... 

You are right that the network is a bridged network, that's the default podman creates on Mac 😄. I've tried to create a non-bridged network, but so far I've failed with that, but I'll try to test it later using an acutal network connection to my homelab server, maybe that will help finding out if it is working

clemensm:
So reading through the Chrome issue tracker that is linked from #327, it seems that the version of Chrome I'm running now is older and still has the `--remote-debugging-address=0.0.0.0` switch. Also it seems that it is not going to come back, so we'll need to use `socat` in the future ....

Issue #336, opened 2025-02-27
Timezone override


CaseyInHaengsin:
Hello,

I have a use case where I need to render pages with different time zones; I found that this is possible via [Emulation.setTimezoneOverride](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setTimezoneOverride). Are you open to a PR that would add support for this functionality?

maltoe:
In the `Navigate` protocol as a `set_timezone` option perhaps? Sure.

Be aware there's a "hidden feature" that allows you to supply custom protocols as well, which would make this task rather easy to do without changes to ChromicPDF

- https://github.com/bitcrowd/chromic_pdf/blob/main/CHANGELOG.md#added
- https://github.com/bitcrowd/chromic_pdf/blob/main/test/integration/custom_protocol_test.exs#L7-L16

but also open to add this as a new option

CaseyInHaengsin:
Thank you, @maltoe! I am open to using the hidden feature if that makes more sense. I did want to attempt implementing the `set_timezone` option. I'm definitely open to feedback/suggestions/teasing on PR #338.

maltoe:
Sorry, never followed up on this. It's merged but we didn't cut a release yet, can do that later.

PR #338, opened 2025-03-03
Timezone override

CaseyInHaengsin:
For #336 
This PR adds a set_timezone option to the navigate protocol.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Added
 
-- :timezone to navigation protocol
+- `print_to_pdf` now supports a `:timezone` option to set the timezone before navigating with `Emulation.setTimezoneOverride`
 
 ## [1.17.0] - 2024-08-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Added
+
+- :timezone to navigation protocol
+
 ## [1.17.0] - 2024-08-09
 
 ### Changed
--- a/test/integration/pdf_generation_test.exs
+++ b/test/integration/pdf_generation_test.exs
@@ -396,7 +396,7 @@ defmodule ChromicPDF.PDFGenerationTest do
 
   describe "timezone can be set thru print_to_pdf/2" do
     setup do
-      start_supervised!({ChromicPDF, offline: false})
+      start_supervised!(ChromicPDF)
       start_supervised!(TestServer.bandit(:http))
 
       %{port: TestServer.port(:http)}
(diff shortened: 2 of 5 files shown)


test/integration/pdf_generation_test.exs
```diff
@@ -394,6 +394,25 @@ defmodule ChromicPDF.PDFGenerationTest do
...
+      start_supervised!(TestServer.bandit(:http))
+
+      %{port: TestServer.port(:http)}
+    end
+
+    @tag :disable_logger
+    test "timezone can be set", %{port: port} do
+      input = {:url, "http://localhost:#{port}/timezone_echo"}
+
+      print_to_pdf(input, [set_timezone: %{timezoneId: "America/New_York"}], fn text ->
```
andreasknoepfle, line 409: Since `timezoneId` seems to be the only parameter of this, why not just simplify the api a bit: 

```suggestion
      print_to_pdf(input, [timezone: "America/New_York"], fn text ->
```

andreasknoepfle (commented):
Can you also prepare an entry for the `CHANGELOG.md`? Like under `[Unreleased]`. You can check the history of the file to see how we normally do it.

test/integration/pdf_generation_test.exs
CaseyInHaengsin: Thank you for that feedback, @andreasknoepfle! I've made that change, and also updated the `CHANGELOG` (thank you for suggesting to look through history for an example).

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Added
+
+- :timezone to navigation protocol
```
andreasknoepfle, line 5:
```suggestion
- `print_to_pdf` now supports a `:timezone` option to set the timezone before navigating with `Emulation.setTimezoneOverride`
```

test/integration/pdf_generation_test.exs
```diff
@@ -394,6 +394,25 @@ defmodule ChromicPDF.PDFGenerationTest do
     end
   end
 
+  describe "timezone can be set thru print_to_pdf/2" do
+    setup do
+      start_supervised!({ChromicPDF, offline: false})
```
maltoe, line 399: Is the `offline: false` necessary for this test?

CaseyInHaengsin: Sorry about that, @maltoe - it is not required. I've removed it!

CHANGELOG.md
CaseyInHaengsin: Thank you for this, @andreasknoepfle! Much clearer; I appreciate it.

PR #341, opened 2025-06-29
Bump ex_doc

srcrip:
see title


maltoe:
Thank you, but what's the cause?

PR #346, opened 2026-01-18
Require Elixir 1.12 to fix CI

maltoe:
Somehow `nimble_parsec` as used in CI caches now requires Elixir 1.12, see https://github.com/bitcrowd/chromic_pdf/actions/runs/21062165943/job/60570795274?pr=345

It's also already been almost 5 years since Elixir 1.12 so I think we're fine to update our minimum Elixir version now.


maltoe:
Removed 1.11 from required checks, but ofc github won't realize this for this branch anymore

andreasknoepfle (approved):

andreasknoepfle:
> Removed 1.11 from required checks, but ofc github won't realize this for this branch anymore

Somehow they were still in there. Removed them now.

maltoe:
Maybe i didn't save :see_no_evil:

